PR #117, opened 2015-01-10
Fixed conditional for the control loop.

auxiliary-character:


virtuald:
Oops! Sorry about that.

Issue #121, opened 2015-01-13
Add setEncoderIndexSource API


virtuald:
https://github.com/robotpy/allwpilib/commit/1b5cabc0d100421a238cac9a13ad1178b2b04c9d

... I think we need to rebuild the HAL binary before we make this change, correct?

PeterJohnson:
Yes, but my buildbot regularly builds binaries (see http://www.tortall.net/~robotpy/hal/) so it's just a matter of updating hal_version in hal-roborio/setup.py at the same time the function is added to hal-base.

computer-whisperer:
I can work on this.

computer-whisperer:
This was finished in #130

PR #130, opened 2015-01-17
Add setEncoderIndexSource to Encoder

computer-whisperer:
What version do we want to upgrade the HAL binary to?

Implements #121.


virtuald:
I think the latest one that's available makes sense.

PR #132, opened 2015-01-18
Can jag free

computer-whisperer:
Added free() to CANJaguar, as is upstream.


virtuald:
Thanks! Sorry I missed this.

PR #135, opened 2015-02-14
Markup the hal_dict, and add a recursive update function

virtuald:
This will make it possible for a simulator to update HAL wholesale without race conditions. I still need to test this more to make sure that this isn't an insane idea, but welcome comments on it.

### DIFF
--- a/hal-sim/hal_impl/data.py
+++ b/hal-sim/hal_impl/data.py
@@ -4,14 +4,24 @@
 
 from hal import constants
 import sys
+import copy
 
 import logging
 logger = logging.getLogger('hal.data')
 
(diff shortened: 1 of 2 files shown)


virtuald:
The basic idea here is that a simulator will have its own copy of the hal_in_data, and just modify the values there, and use update_hal_data to recursively merge in the data.

This doesn't make a lot of sense if you're running locally (such as the gazebo simulator) -- just update hal_data directly. However, if you're running over a web connection, sending just the in data via JSON should be a big win.

sarosenb:
Would this help with the issues that cause us to hang in command based?

virtuald:
Nope

PR #144, opened 2015-02-25
Don't throw errors when HAL emits a warning

### LINKED ISSUE #143: HAL STATUSFUNC should distinguish between warnings and errors
This bug happens particularly when accessing the PDP. See http://www.chiefdelphi.com/forums/showthread.php?t=135145 for details.

virtuald:
- Fixes #143

To test this, you can do the following:

```
git clone https://github.com/robotpy/robotpy-wpilib.git
cd robotpy-wpilib
git checkout hal-warning
```

On a linux/OSX system, or if you have MSYS installed on windows, you can run `devtools/build_and_install.sh`. If you don't have either of those, post a comment and I can write something up that will get you going.

### DIFF
--- a/hal-base/hal/functions.py
+++ b/hal-base/hal/functions.py
@@ -41,7 +41,7 @@ def _STATUSFUNC(name, restype, *params, out=None, library=_dll,
         elif status.value < 0:
             raise HALError(getHALErrorMessage(status.value))
         elif status.value > 0:
-            warnings.warn(getHALErrorMessage(status.value))
+            warnings.warn(getHALErrorMessage(status.value), stacklevel=2)
             return rv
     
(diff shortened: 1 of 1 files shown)


aerobit:
Seems to have worked for me! I deployed it and it's not crashing anymore. On the other hand, I'm not getting any HAL warnings, so the problem might just have fixed itself (CAN is weird sometimes)

aerobit:
Even better! Just got a warning.

    /usr/local/lib/python3.4/site-packages/wpilib/powerdistributionpanel.py:56: UserWarning: CTRE CAN Recieve Timeout
      return hal.getPDPTotalCurrent()

Works perfectly, thanks for the patch :)

PR #153, opened 2015-10-16
LiveWindow work: Fix DS thread crashing from exceptions thrown by LV device hooks.

### LINKED ISSUE #148: Running LiveWindow after calling free() on a resource results in "operation on freed port" errors.
There doesn't seem to be a mechanism for removing components from LiveWindow after they have been freed. I'm not quite sure how this should be handled. 

One way would be to add a ```removeComponent()``` method to LiveWindow and have every ```free()``` method call it. Another way (possibly a good idea to do in any case) would be to have LiveWindow catch and log exceptions that it receives from the device objects.

@virtuald, @PeterJohnson what do you think?

computer-whisperer:
Added a try/except to catch any exceptions thrown by device's `.startLiveWindowMode()` hook.

Also added a `LiveView.removeComponent()` method, and now call it from the `free()` method of every device that has a `LiveWindow.addDevice()` in it's `__init__()`.

This fixes #148

### DIFF
--- a/wpilib/wpilib/cantalon.py
+++ b/wpilib/wpilib/cantalon.py
@@ -168,6 +168,7 @@ class CANTalon(LiveWindowSendable, MotorSafety):
         return self._handle
 
     def free(self):
+        LiveWindow.removeComponent(self)
         self._handle_finalizer()
 
     def pidWrite(self, output):
--- a/wpilib/tests/conftest.py
+++ b/wpilib/tests/conftest.py
@@ -67,7 +67,13 @@ def wpilib(_module_patch, hal, hal_data):
     """Actual wpilib implementation"""
     import wpilib
     return wpilib
-    
+
+@pytest.fixture(scope="function")
+def networktables():
+    """Networktables instance"""
+    import networktables
+    networktables.NetworkTable.setTestMode()
+    return networktables
 
 #
 # Mock fixtures for testing things that don't have to interact with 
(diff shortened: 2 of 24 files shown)


wpilib/wpilib/cantalon.py
```diff
@@ -168,6 +168,7 @@ def handle(self):
         return self._handle
 
     def free(self):
+        LiveWindow.removeComponent(self)
```
virtuald: Seems like this should be done in the finalizer?

virtuald:
If you can fix the tests, that would be excellent. Other than that, this seems ok to me.

wpilib/wpilib/cantalon.py
computer-whisperer: I don't think so, because LiveWindow components are stored in a dictionary with the device objects as the keys of the dictionary. You need to have the device object in order to remove the entry from that dictionary, and the finalizer doesn't have access to that object, since it is supposed to run after the object has been deleted.

computer-whisperer:
Tests fixed.

computer-whisperer:
Rebased

wpilib/wpilib/cantalon.py
virtuald: I suppose that's true.

virtuald: In fact, if the dictionary is holding references to the object, then the finalizer will never run anyways. Perhaps it should be a dictionary of weak references.

PR #154, opened 2015-10-17
Adjusting for upstream HAL API updates, needs more work.

computer-whisperer:
This is not ready to merge yet, I just needed a place to discuss stuff with the group.

Also, feel free to double check any of my work.

### DIFF
--- a/hal-base/hal/functions.py
+++ b/hal-base/hal/functions.py
@@ -15,7 +15,7 @@ def hal_wrapper(f):
     '''
     
     wrapped = globals()['_' + f.__name__]
-    if hasattr(f, 'fndata'):
+    if hasattr(wrapped, 'fndata'):
         f.fndata = wrapped.fndata
     return f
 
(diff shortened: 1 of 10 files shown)


hal-base/hal/functions.py
```diff
@@ -494,14 +484,14 @@ def cleanNotifier(notifier):
 #############################################################################
 # PDP
 #############################################################################
-getPDPTemperature = _STATUSFUNC("getPDPTemperature", C.c_double)
```
computer-whisperer: An Issue I am running into is that on some functions, like these PDP functions, the "status" argument is not the final argument in the c++ API, like we expect. I suspect someone got lazy when they added arguments to these, and didn't rearrange them to match the rest of the API. See https://github.com/robotpy/allwpilib/blob/master/hal/include/HAL/PDP.hpp

virtuald: We should get them to fix that, I don't want to do any extra work.

computer-whisperer:
Also @PeterJohnson fyi the folders at http://www.tortall.net/~robotpy/hal/ seem to all be empty for builds newer than jenkins-stable-2015.340-112-g4e46692.

hal-base/hal/functions.py
virtuald: I've pushed a [changeset](https://usfirst.collab.net/gerrit/#/c/1085/) into gerrit to get this fixed, hopefully should get merged pretty quickly.

virtuald: See the [halfix](https://github.com/robotpy/allwpilib/blob/halfix/hal/include/HAL/PDP.hpp) branch in allwpilib.

computer-whisperer: Don't you need to fix the callers of these functions in both the java and c++ wpilib? Not just the HAL api?

virtuald: Yes, look at the rest of the commit. ;)

computer-whisperer: Ok, my bad.

computer-whisperer:
I was able to run this successfully in both the simulation environment and a real roborio. Unless you see things that need to be changed, I think this is ready to merge. Only caveat is that the PDP functions will be broken until Dustin's change gets merged upstream.

PeterJohnson:
Builds should be working again as of 2015.340-176.  Note I've changed the build to use gcc 4.8.2 so it will work out of the box with the 2016 image, but this may break things on the 2015 image.

computer-whisperer:
Thanks!

PR #163, opened 2015-12-28
Various updates to installer.

computer-whisperer:
Added commands for downloading and installing opkg files, updated architecture and repo link for 2016.

### DIFF
--- a/installer/installer.py
+++ b/installer/installer.py
@@ -546,14 +546,18 @@ class RobotpyInstaller(object):
     opkg_cache = abspath(join(dirname(__file__), 'opkg_cache'))
     
     # opkg feed
-    opkg_feed = 'http://www.tortall.net/~robotpy/feeds/2014/'
-    opkg_arch = 'armv7a-vfp-neon'
+    opkg_feed = 'http://www.tortall.net/~robotpy/feeds/2016/'
+    opkg_arch = 'cortexa9-vfpv3'
     
     commands = [
         'install-robotpy',
         'download-robotpy',
         'install',
-        'download'
+        'download',
+        'install-pip',
+        'download-pip',
+        'install-opkg',
+        'download-opkg'
     ]
 
     def __init__(self):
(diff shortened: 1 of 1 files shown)


computer-whisperer:
If this is merged I'll pr it to pyfrc.

virtuald:
This is great... but it doesn't appear to resolve dependencies?

computer-whisperer:
Not for opkgs at the moment. I could look into doing that.

virtuald:
I think it should be pretty straightforward to do now that you've added arbitrary opkg installation. The dependency data is in the Packages file, which we already download.

virtuald:
Do you think you'll have time this week to do the dependency resolving?

computer-whisperer:
I can try to put some time into it tomorrow. It shouldn't be that hard.

computer-whisperer:
Finished, it can now resolve, download, and install dependencies -- including from the base NI repo.

virtuald:
Awesome, I'll try it out tonight. The only nit I have is that I think you broke python 2.7 compatibility with the addition of the urllib import, but that should be easy enough to fix. 

Only reason I care about python 2.7 compatibility is I would like to position this to be used for general opkg installation, and it makes the barrier for entry a bit lower.

installer/installer.py
```diff
@@ -546,14 +564,18 @@ class RobotpyInstaller(object):
     opkg_cache = abspath(join(dirname(__file__), 'opkg_cache'))
     
     # opkg feed
-    opkg_feed = 'http://www.tortall.net/~robotpy/feeds/2014/'
-    opkg_arch = 'armv7a-vfp-neon'
+    opkg_feed = 'http://www.tortall.net/~robotpy/feeds/2016/'
```
virtuald: This doesn't seem to be used anymore, since you're hardcoding it below?

computer-whisperer:
What urllib import? I didn't add any.

installer/installer.py
computer-whisperer: Right, I'll fix this soon.

installer/installer.py
```diff
@@ -30,7 +31,7 @@
 
 from collections import OrderedDict
 from distutils.version import LooseVersion
-from urllib.request import urlopen, urlretrieve
```
virtuald: Oh, ha. I bet I added this. Ignore me.

virtuald:
I like it, seems to work. I'm going to add a blacklist for system packages like libc6 though, as they aren't necessary and just eat up more disk space.

In the future I wonder if we should be smarter about the files that we copy via opkg, and if we should clean the cache on the roborio side. They can take up quite a bit of space.

PR #186, opened 2016-02-25
Fix thread errors in PIDController.

james-ward:
Due to asynchronous calls to Timer.delay() it was possible for getDeltaSetpoint()
to throw a divide by zero error.
Cancelling the TimerTask thread outside the mutex avoids some deadlocks,
which may be the reason behind the testing hangs discussed in
robotpy/pyfrc#46

I have run my team's test code and the NavX example that cause the Travis failure repeatedly and they do not hang. I was able to intermittently get the hang before I made these changes.


virtuald:
This looks great, thanks so much!

PR #187, opened 2016-03-06
Added warning to not use during official events in the readme.

avecchio:
tldr
Team used unofficial language at an event. Advise that this project should not be used at competitions.

So I was a CSA at an event and was not required to assist a team that was experiencing issued with their robot since they were using this library. I find this project to be awesome and a fun way to learn software outside of official FIRST events. But seeing as how python does not cooperate well with embedded systems, it may not be fully compatible running on a roborio. So I added a warning into the readme so this can be avoided in the future. If you wish to ignore my warning (as it is kind of a rant), please do so. This PR was issued right after a competition.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,16 @@
 RobotPy WPILib
 ==============
 
+WARNING: This library is not officially supported by FIRST and therefore should be advised not to be used in competitions.
+Since python is not supported, Control System Advisors and Field Technical Adivsors are not trained and not required to assist
+if a non official language is used on a robot during an official FIRST event.
+The officially supported languages can be found here: [FRC Control Systems](https://wpilib.screenstepslive.com/s/4485)
+This project can be used in any manner outside of an official FIRST Robotics Event.
+
 [![Build Status](https://travis-ci.org/robotpy/robotpy-wpilib.svg)](https://travis-ci.org/robotpy/robotpy-wpilib)
 
-This repository contains the source code for a 100% python implementation of WPILib, 
-the library used to interface with hardware for the FIRST Robotics Competition. 
+This repository contains the source code for a 100% python implementation of WPILib,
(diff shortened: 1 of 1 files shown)


PeterJohnson:
Your change as proposed is a little too strongly worded, particularly the statement "This project can be used in any manner outside of an official FIRST Robotics Event." -- as this is emphatically NOT the case; there are teams which have successfully used this project up to the championship level, and there is nothing in the rules that restricts languages/libraries to only the officially supported ones.

However, I agree that a warning that this is not an officially supported language/library for competition would be good to make more prominent in the README and on the website.  Cautionary statements along these lines were quite prominently placed in the original project announcements, but have since been buried a bit in the FAQ of the website (http://robotpy.github.io/faq/, Competition section), and as such there may be teams unaware of the full risks involved of using an unofficial language for competition.  I would think placing a combination of the text in the FAQ and a slightly softened version of your proposed change in the README (and increasing its prominence on the website) would be sufficient to address this concern.

Note for other robotpy developers: concerns such as the one raised here do need to be addressed, or there is a risk that FIRST might limit language options in the rules in future years.  More prominent caution about this language being officially unsupported may help teams make the right language choice for their technical abilities and risk tolerance.

virtuald:
@avecchio other than language syntax, what issue was the team having? And, what team number, just out of curiousity?

In previous years I mentioned during CSA meetings that teams can contact me directly for Python support if required, but the telecons did not happen this year.

virtuald:
I've added notices to the documentation site and the README that should address your concerns. See 1b59b163a1851487c7f99e6c1a9c5393598f5080 for more details.

virtuald:
FWIW, @frc1418 (former team of mine) won both of their district events this year, and they are using RobotPy.

Twinters007:
And captained an alliance on Newton!

PR #191, opened 2016-03-29
Copy installer fix from pyfrc

computer-whisperer:
Matches pyfrc#53


installer/installer.py
```diff
@@ -460,7 +460,8 @@ def sftp(self, src, dst, mkdir=True):
                                     first = False
                                 else:
                                     rd = join(rdst, relpath(d, src))
-                                    fp.write('mkdir "%s"\n' % rd)
```
virtuald: Hm, why is this conditional?

PR #194, opened 2016-05-04
Fixed the IPv6 deploy problem with unpacking socket.getaddrinfo().

ryannazaretian:
Like the pyfrc one, this is untested, but I think it should work.


vanjac:
This has been causing problems for our team members. I've tested this fix on all computers with the problem and it seems to work. Could we have this merged?

virtuald:
This fix only applies to the robotpy installer. This was merged into pyfrc (for deploying robot code) in https://github.com/robotpy/pyfrc/pull/57, if you update pyfrc you should find that the problem will go away.

PR #198, opened 2016-08-27
Fix internal "read" method name

FRC2539:
Based on line 85 and by comparison with the _write method, this appears to be a typo.

### DIFF
--- a/wpilib/wpilib/preferences.py
+++ b/wpilib/wpilib/preferences.py
@@ -340,7 +340,7 @@ class Preferences:
             except ImportError:
                 pass
 
-    def read(self):
+    def _read(self):
         """The internal method to read from a file. This will be called in its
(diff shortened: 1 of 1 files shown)


virtuald:
If I were to accept this, you would need to update line 85 also, which calls read. 

`_` is a prefix often used to denote internal methods that you shouldn't call unless you have a good reason. What's your reason?

FRC2539:
Are we looking at the same line 85?

It looks like line 85, in all public branches of this repository currently reads:

> reader = threading.Thread(target=self._read,

That already has the underscore before "read", which is why this change is necessary. Without the change, python throws the following exception when trying to use Preferences:

> AttributeError: 'Preferences' object has no attribute '_read'

virtuald:
Oh ha, I read the diff backwards. Busy weekend.

Issue #195, opened 2016-05-08
Deprecate command based programming in robotpy


virtuald:
Bottom line: it sucks (in robotpy) and isn't pythonic, and I don't want to spend time to make it better

Creating a command based robot in python is clunky and leads to all sorts of problems. With the addition of state machine support, magicbot is more pythonic, more powerful, and easier to use. 

I would propose using the warnings.warn mechanism to annoy users when they used the various command based components, and to point at a HTML page that explains why and alternatives.

FRC2539:
We are just starting with Python, so perhaps our viewpoint will change in the future. Our first program, done only a month ago, was to modify the example command based program to drive a single motor on our prototyping system. We're pleased to say it worked well. So well, in fact, that we decided to try porting last season's program, which is C++ and command based. However, we have not gotten a chance to drive with it yet, though we hope to soon. (We're only meeting once a week during the off-season.)

Our team actually tried Python for Rebound Rumble but at that time they couldn't even get the program to start, so they went with C++. However, Python is now looking like a first-class citizen of the FRC ecosystem and many of the team programmers would rather use it. That being said, we have several years of experience with the command based paradigm and porting to something like magicbot is daunting. Here are a few use cases we have that we don't see an easy way to implement with magicbot, followed by few things we think mitigate the arguments against command based.

If there are existing solutions to the issues we've been puzzled by, listing them in the proposed warning message HTML page would probably help other teams looking to port their programs to Python.

# Command-Based Tricks
In no particular order

## Controller Flexibility
One of the really great things about the command based paradigms is that it makes controller coupling easy and flexible. The syntax for assigning an action to a button is just `button.event(Command())`. By contrast, the example given for magicbot is five lines:

```
try:
    if self.joystick.getTrigger():
        self.component.doSomething()
except:
    self.onException()
```

We have gotten into the habit of creating multiple controller layouts for different situations. For example, during development we just assign every last button on a controller to something, because it's fast for testing. At competitions we generally have two controllers (with two drivers), and we try to use buttons that are far away from each other to avoid accidentally triggering the wrong action (a lesson learned the hard way, sadly). For open houses we only want visitors to be able to drive the robot (slowly) and do a few simple actions that can't cause any damage.

With C++ we could include different files for each occasion. With Python we plan to run different functions in `OI` depending on the situation. To do the same with magic bot it looks like we'd either need to have conditional logic inside of `teleopPeriodic()` or assign different functions to `teleopPeriodic`.

[...]

PR #199, opened 2016-08-27
Use "append" for NetworkTable Arrays

FRC2539:
StringArray and NumberArray objects do not have an "add" method, but they do have "append" which appears to have the desired behavior.

### DIFF
--- a/wpilib/wpilib/command/scheduler.py
+++ b/wpilib/wpilib/command/scheduler.py
@@ -244,8 +244,8 @@ class Scheduler(Sendable):
             self.ids.clear()
             # Set the the running commands
             for command in self.commandTable:
-                self.commands.add(command.getName())
-                self.ids.add(id(command))
+                self.commands.append(command.getName())
+                self.ids.append(id(command))
             self.table.putValue("Names", self.commands)
             self.table.putValue("Ids", self.ids)
 


virtuald:
If you're exploring the command framework, take note of https://github.com/robotpy/robotpy-wpilib/issues/195

FRC2539:
We have taken note of it. Say a team was looking to use the command-based paradigm in what they feel is a pythonic manner in order to maintain consistency with the other languages (and hopefully therefore be able to help and get help from teams that use other languages at competitions). What would they need to do in order to avoid it being deprecated?

james-ward:
As someone who spent a bit of time working on the command based examples and getting command based stuff to work better with the simulator at the end of last year, I feel qualified to say "don't do command based!"
We started using it this year, and as soon as Dustin released his MagicBot framework we switched over. Even refactoring in the middle of the build season was worth it. It is so much easier and cleaner to work with.
I would prioritise ease of use, lack (less?? :-P ) weird edge cases, and speed of design iteration above "compatibility" with other languages. Using Python is gaining traction, but you are still likely to come across FTAs that won't help when they find you are using Python (they might even tell you that using Python on the bot is not legal... fun times...), so consistency with other languages is unlikely to get you much there anyway. [For the sake of those reading this post in the future, using Python on your robot at a competition absolutely is legal (and highly recommended!)]

virtuald:
If you have a good rationale for keeping it, then a comment on #195 would be appropriate. Given that the goal of robotpy's WPILib port is to make everything available in Java available in Python also, it won't go away for the moment -- but if we add the warnings then you'll just get annoying warnings if you use it, and I'm not going to put much effort into improving it beyond what Java provides.

Fundamentally, I dislike Command-based programming because it requires too much boilerplate. It takes a large amount of text to do simple things -- after all, this is one of the reasons why RobotBuilder exists. It has a very 'Java' feel to it, and I don't think you can change it to be pythonic without significant changes.

One last complaint is that Python code tends to crash very easily as compared to Java (for example, accessing an attribute that doesn't exist, which would get caught by the Java compiler). The command framework doesn't currently have a good way to deal with these sorts of things, as Java doesn't have this problem. There are probably ways of dealing with this, but I haven't dug into it very much.

PR #203, opened 2016-10-26
Fix ValueError

FRC2539:
The partition method returns a 3-tuple, so we need three variables to unpack it.


virtuald:
FYI, I think the preferences class will get totally rewritten now that pynetworktables supports NT3 (95% done) which has persistent variables.

PR #214, opened 2016-11-17
Update smartdashboard to 2017 changes

ArthurAllshire:
@virtuald let me know if there is anything that you want done differently with this, especially with the putType methods.


virtuald:
This seems good to me, the only thing I might recommend is adding documentation to the things that have flags... probably should document where one can find the flags.

virtuald:
I wouldn't be opposed to converting them all to classmethods too, but I think it's fine.

PR #215, opened 2016-11-20
Update command classes for 2017

FRC2539:
Since we were the ones that got [these changes](https://github.com/wpilibsuite/allwpilib/pull/238) merged into WPILib, it seemed like our responsibility to handle the porting for Python. We are not a beta team, so we can't test with the new HAL, but the code works with last year's firmware and there are no low-level changes.


virtuald:
Thanks!

PR #218, opened 2016-11-27
newcontroldata needs to be updated

ariovistus:
or teleopPeriodic never gets called in iterative robot.

at least with frcsim


virtuald:
You're correct that there is a bug, but that isn't where that variable needs to be updated. Since this source code is a translation of the java source code, unless there's a compelling reason I try to imitate that. The variable needs to be set in _run, not _getData.

I'll push a fix momentarily.

PR #219, opened 2016-11-29
Fix javadoc converter to exclude argument types

ArthurAllshire:
Didnt use any fancy regex stuff, but the fix works.


ArthurAllshire:
I've just realized that this breaks with multiple arguments - working on a fix

virtuald:
Thanks!

PR #226, opened 2016-12-21
Update 2017

james-ward:
Some fairly trivial updates.


wpilib/wpilib/spi.py
```diff
@@ -1,4 +1,4 @@
-# validated: 2015-12-30 DS 4b04073 athena/java/edu/wpi/first/wpilibj/SPI.java
```
virtuald: All of the HAL API names changed, at a minimum those need to be updated.

james-ward: Sorry, didn't realise that was something that needed doing. Just read your commit about HAL stuff - I see that we need to check for that.
Test for SPI passes, but ultrasonic fail because it calls the failing `digitalsource.py` file.

PR #231, opened 2016-12-24
Update I2C and solenoids

james-ward:


virtuald:
Looks like 8c4ad6242298c97993e472314ae78901deffd2c0 removed allocations stuff in SolenoidBase, which were omitted from this PR. I wonder if git-source-track messed up the source logging somehow?

james-ward:
Honestly, I don't think this was git-source-track's fault - it was operator error.
Mea culpa.

PR #234, opened 2016-12-27
Update DigitalSource, DigitalInput, Digital Output

james-ward:


james-ward:
There are still a number of unit tests that aren't passing, but that's because of the hal-sim implementations.

PR #235, opened 2016-12-29
Hal for tests

james-ward:
Every unit test runs individually except test_driverstation.py.

There is still a resource locking issue when running all tests together.

### DIFF
--- a/hal-sim/hal_impl/functions.py
+++ b/hal-sim/hal_impl/functions.py
@@ -818,23 +818,23 @@ def freeDIOPort(dioPortHandle):
 
 def allocateDigitalPWM(status):
     status.value = 0
-    assert False
+    return types.DigitalPWMHandle()
 
 def freeDigitalPWM(pwmGenerator, status):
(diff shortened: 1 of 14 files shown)


hal-sim/hal_impl/functions.py
```diff
@@ -1475,23 +1482,51 @@ def getUserCurrentFaults3V3(status):
...
+    channel, fwd = _handle_to_channel(relayPortHandle)
+    if fwd:
+        hal_data['relay'][channel]['fwd'] = False
+    else:
+        hal_data['relay'][channel]['rev'] = False
 
+    
 def checkRelayChannel(channel):
-    assert False
+    return 0 <= channel and channel < kNumRelayHeaders
```
auscompgeek: Sometimes I wonder whether we should just use chained comparisons in these check functions…

virtuald: What do you mean by chained comparisons?

PR #250, opened 2017-01-12
Update NetworkTables usages in Scheduler

FRC2539:
Fixes the ImportError when trying to use the Scheduler, and changes StringArray and NumberArray to Python lists.


virtuald:
Sorry about that! I'm surprised that the command-based robot example worked... or is that because this only happens in test mode?

FRC2539:
The example program doesn't try to send the `Scheduler` to the SmartDashboard, so nothing triggers `initTable()`.

PR #257, opened 2017-01-31
Clean up code here and there

Twinters007:
@virtuald Carter, Erik and I are wondering how you feel about a bit of spring cleaning


virtuald:
This seems like a bad idea.

virtuald:
The only way I would maybe approve this is if you added tests to travis that ensured people didn't mess it up.

Twinters007:
Guess what 2017-robot has

PR #258, opened 2017-02-04
Fixed the turning problem with mecanumDrive_Cartesian

Kainoad:
Instead of turning, our robot spun wheels against itself and got nowhere. The front and rear wheels were spinning opposite each other in such a way that it didn't move sideways either. We determined that for our robot, the front right and rear left wheels were the ones spinning the wrong way. We also checked and made sure that our motors and motor controllers were all connected properly. So far unsure if others had this problem, but this was the case with our robot.


virtuald:
That does not match the code in the Java WPILib. See https://github.com/wpilibsuite/allwpilib/blob/master/wpilibj/src/athena/java/edu/wpi/first/wpilibj/RobotDrive.java#L478 . Your motors are incorrect.

PR #260, opened 2017-02-11
Fixes Conditional Command

FRC2539:
Call the parent's constructor before attempting to set requires.


virtuald:
One of these days we'll have unit tests for the command stuff...

FRC2539:
We don't have a good grasp on the testing framework, but we can add it to our list of things to look at over the summer if we have time.

Any chance this will be merged and in a release before tomorrow night?

virtuald:
Sorry, I thought I had done that. I'll push up a release in a few minutes.

virtuald:
Also, if you find you need to make local modifications, it's pretty easy to deploy a locally modified version RobotPy. Just run `devtools/build_and_deploy.sh`

Issue #253, opened 2017-01-15
Cannot use a relay initialized as Direction::kReverse


virtuald:
Upstream bug at https://github.com/wpilibsuite/allwpilib/issues/458, manifests itself as a TypeError in python, because we cannot pass None to the HAL API. Simulation is currently incorrect, as it detects the None and treats it as a normal condition.

auscompgeek:
Bug has been fixed upstream.

PR #273, opened 2017-07-27
Update Relay to comply with wpilib changes. Fixes 253

Twinters007:


virtuald:
Presumably this won't work without a corresponding HAL change?

Twinters007:
No idea...I also didn't have to change any tests, so I may not understand the underlying issue very well.

Twinters007:
Well, now that I look at the code, I can't tell what has changed between the two, except for the order of the if statements. The current code returns the same results it seems.

virtuald:
If this is up to date, then can you validate it using git source-track too?

Twinters007:
Yeah. Although I don't know how to write tests that test for...whatever was wrong? Maybe you can give some guidance to the problem and I can figure out a solution?

auscompgeek:
I think #253 (and the corresponding upstream issue/PR) should explain it well enough…

Twinters007:
Shouldn't need a HAL change. It's just the get() method that changed upstream as well.

Twinters007:
Should I close this in favor of #304?

auscompgeek:
Given that #304 has NT updates as well, I think that would make sense.

PR #274, opened 2017-08-11
Report uncaught exceptions to DS

Twinters007:
So I think this will work, but I'm not completely sure about the inner workings of wpilib to know if this is the right place to put this. 

I also don't know how to test for this, since it involves reporting to the DS

### DIFF
--- a/wpilib/wpilib/_impl/main.py
+++ b/wpilib/wpilib/_impl/main.py
@@ -14,7 +14,6 @@ from wpilib.driverstation import DriverStation
 def my_except_hook(exctype, value, traceback):
     DriverStation.reportError("ERROR Unhandled {}".format(exctype), True)
     sys.__excepthook__(exctype, value, traceback)
-sys.excepthook = my_except_hook     
 
 def _log_versions():
     import wpilib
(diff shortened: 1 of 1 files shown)


wpilib/wpilib/_impl/main.py
```diff
@@ -9,6 +9,12 @@
 from .logconfig import configure_logging
 import hal_impl
 
+from wpilib.driverstation import DriverStation
+
+def my_except_hook(exctype, value, traceback):
+    DriverStation.reportError("ERROR Unhandled {}".format(exctype), True)
+    sys.__excepthook__(exctype, value, traceback)
+sys.excepthook = my_except_hook
```
virtuald: Hm, seems like you would want to do this only when the code is actually launched, instead of always.

Twinters007: you mean if 'run' is the entry point or whatever?

virtuald: Well. Maybe only when the user's code is about to be launched?

Twinters007:
How about now @virtuald

virtuald:
Turns out, this isn't actually useful:

* Our main thread always catches exceptions in RobotBase
* `sys.excepthook` doesn't catch exceptions that arise in other threads (and this is a particularly annoying problem to solve -- see https://bugs.python.org/issue1230540)

PR #277, opened 2017-09-22
Update adx classes for 2018

ArthurAllshire:


wpilib/wpilib/adxl345_i2c.py
```diff
@@ -150,9 +150,22 @@ def getAccelerations(self):
     def getSmartDashboardType(self):
         return "3AxisAccelerometer"
 
+    def initTable(self, subtable):
+        if subtable is not None:
+            self.xEntry = subtable.getEntry("X")
```
virtuald: In retrospect, pretty sure `getEntry` isn't defined, that's a new ntcore thing. We should probably have some minimal test(s) that create all of the objects and add them to LiveWindow so this sort of thing gets exercised..

PR #280, opened 2017-09-24
Validate most PID files and remove pidinterface

ArthurAllshire:
Still need to finish pidcontroller - but can't do that until we figure out the NT stuff.


ArthurAllshire:
Oops - forgot to remove the import of pidinterface in interfaces/__init__.py, fixed.

PR #282, opened 2017-09-29
Update hal functions to match headers

Twinters007:


Twinters007:
It looks like this was the only thing to update...

virtuald:
But @ArthurAllshire said there were functions he was missing?

Twinters007:
he did say that didn't he...hmmm

hal-sim/hal_impl/functions.py
```diff
@@ -1023,8 +1035,7 @@ def getMatchTime(status):
         return (hooks.getFPGATime() - hal_data['time']['match_start'])/1000000.0
 
 def waitForDSData():
-    with hooks.ds_cond:
-        hooks.ds_cond.wait()
+    waitForDSDataTimeout(0)
```
virtuald: I think this should be `None`

hal-sim/hal_impl/functions.py
```diff
@@ -243,6 +243,12 @@ def getBrownedOut(status):
     status.value = 0
     return False
 
+def baseInitialize(status):
```
virtuald: Hm, shouldn't this call initialize()? Why are there two of them now?

hal-sim/hal_impl/functions.py
```diff
@@ -1044,6 +1055,19 @@ def observeUserProgramTeleop():
 def observeUserProgramTest():
     hal_data['user_program_state'] = 'test'
 
+def releaseDSMutex():
```
virtuald: This is a bit strange, why isn't there a corresponding acquire function? What is the original function supposed to do (did you read the original source?)

hal-sim/hal_impl/functions.py
```diff
@@ -1044,6 +1055,19 @@ def observeUserProgramTeleop():
...
+    hooks.ds_cond.release()
+
+def isNewControlData():
+    with hooks.ds_cond:
+        return True
```
virtuald: This also seems wrong.

hal-sim/hal_impl/functions.py
Twinters007: the variable should be?

virtuald: 0 should be None, as None is an infinite timeout... I don't know if 0 is.

hal-sim/hal_impl/functions.py
Twinters007: Couldn't tell you...

hal-sim/hal_impl/functions.py
```diff
@@ -9,6 +9,7 @@
 from hal_impl.sim_hooks import SimHooks
 
 import logging
+from hal_impl.mode_helpers import notify_new_ds_data
```
auscompgeek: This creates an import loop. This completely breaks on Python 3.4.

Twinters007: Ahh that’s why that broke

virtuald:
Looks like the headers/libraries are being pushed again: http://first.wpi.edu/FRC/roborio/maven/development/edu/wpi/first/hal/hal/2018.1.1-beta-1-20171018005731-8-g2fc6068/

There's a separate file for headers and for the library though, so we'll have to update the code that downloads them now.

virtuald:
I'm going to grab this PR and do it today sometime I think, since I suspect it'll end up blocking other PRs.

virtuald:
Closed in favor of #294

PR #283, opened 2017-10-02
getJoystickName shouldn't bomb out

ariovistus:
also, add a slot for joystick name


Twinters007:
Bomb out isn't a very technical term. Can you elaborate?

PR #284, opened 2017-10-03
update solenoid and solenoidbase for 2018

ariovistus:
complete with calls to nonexistent networktables api

### DIFF
--- a/wpilib/tests/test_pwm.py
+++ b/wpilib/tests/test_pwm.py
@@ -151,7 +151,7 @@ def test_pwm_updateTable(pwm):
 
 def test_pwm_valueChanged(pwm):
     pwm.setSpeed = MagicMock()
-    pwm.valueChanged(_EntryNotification(name='Value', value=0.5, flags=1, local_id=1))
+    pwm.valueChanged(None, None, 0.5, None)
     pwm.setSpeed.assert_called_once_with(0.5)
     
 def test_pwm_startLiveWindowMode(pwm, pwm_data):
--- a/wpilib/tests/test_pwm.py
+++ b/wpilib/tests/test_pwm.py
@@ -1,6 +1,5 @@
 import pytest
 from unittest.mock import MagicMock
-from ntcore.entry_notifier import _EntryNotification
 
 #
 # Module-specific fixtures
--- a/wpilib/tests/test_pwm.py
+++ b/wpilib/tests/test_pwm.py
@@ -1,5 +1,6 @@
 import pytest
 from unittest.mock import MagicMock
+from ntcore.entry_notifier import _EntryNotification
 
 #
 # Module-specific fixtures
@@ -137,20 +138,20 @@ def test_pwm_getSmartDashboardType(pwm):
     assert pwm.getSmartDashboardType() == "Speed Controller"
 
 def test_pwm_updateTable(pwm):
-    pwm.getTable = MagicMock()
+    pwm.valueEntry = MagicMock()
     pwm.getSpeed = MagicMock()
     # normal case
     pwm.updateTable()
-    pwm.getTable.return_value.putNumber.assert_called_once_with("Value", pwm.getSpeed.return_value)
+    pwm.valueEntry.setDouble.assert_called_once_with(pwm.getSpeed.return_value)
     # None case
     pwm.getSpeed.reset_mock()
-    pwm.getTable.return_value = None
+    pwm.valueEntry = None
     pwm.updateTable()
     assert not pwm.getSpeed.called
 
 def test_pwm_valueChanged(pwm):
     pwm.setSpeed = MagicMock()
-    pwm.valueChanged(None, None, 0.5, None)
+    pwm.valueChanged(_EntryNotification(name='Value', value=0.5, flags=1, local_id=1))
     pwm.setSpeed.assert_called_once_with(0.5)
     
 def test_pwm_startLiveWindowMode(pwm, pwm_data):
--- a/wpilib/tests/test_relay.py
+++ b/wpilib/tests/test_relay.py
@@ -1,5 +1,6 @@
 import pytest
 from unittest.mock import MagicMock, call
+from ntcore.entry_notifier import _EntryNotification
 
 
 @pytest.fixture(scope="function")
@@ -185,22 +186,22 @@ def test_relay_getSmartDashboardType(relay):
 
 @pytest.mark.parametrize("value", [None, "Off", "On", "Forward", "Reverse"])
 def test_relay_updateTable(value, relay):
-    relay.getTable = MagicMock()
+    relay.valueEntry = MagicMock()
     relay.get = MagicMock()
     if value is None:
-        relay.getTable.return_value = None
+        relay.valueEntry = None
         relay.updateTable()
         assert not relay.get.called
     else:
         relay.get.return_value = getattr(relay.Value, "k"+value)
         relay.updateTable()
-        relay.getTable.return_value.putString.assert_called_once_with("Value", value)
+        relay.valueEntry.setString.assert_called_once_with(value)
 
 @pytest.mark.parametrize("value", ["inv", "Off", "On", "Forward", "Reverse"])
 def test_relay_valueChanged(value, relay):
     relay.set = MagicMock()
-    relay.valueChanged(None, None, value, None)
+    relay.valueChanged(_EntryNotification(name='', value=value, flags=1, local_id=1))
     if value == "inv":
-        assert not relay.set.called
+        relay.set.assert_called_once_with(getattr(relay.Value, "kOff"))
     else:
         relay.set.assert_called_once_with(getattr(relay.Value, "k"+value))
--- a/wpilib/tests/test_relay.py
+++ b/wpilib/tests/test_relay.py
@@ -1,6 +1,5 @@
 import pytest
 from unittest.mock import MagicMock, call
-from ntcore.entry_notifier import _EntryNotification
 
 
 @pytest.fixture(scope="function")
--- a/wpilib/tests/test_relay.py
+++ b/wpilib/tests/test_relay.py
@@ -200,7 +200,7 @@ def test_relay_updateTable(value, relay):
 @pytest.mark.parametrize("value", ["inv", "Off", "On", "Forward", "Reverse"])
 def test_relay_valueChanged(value, relay):
     relay.set = MagicMock()
-    relay.valueChanged(_EntryNotification(name='', value=value, flags=1, local_id=1))
+    relay.valueChanged(None, None, value, None)
     if value == "inv":
         relay.set.assert_called_once_with(getattr(relay.Value, "kOff"))
     else:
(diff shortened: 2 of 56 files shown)


virtuald:
They'll exist soon.

wpilib/wpilib/solenoidbase.py
```diff
@@ -23,13 +23,36 @@ def __init__(self, moduleNumber):
...
+        return hal.getAllSolenoids(moduleNumber)
+
     def getAll(self):
         """Read all 8 solenoids from the module used by this solenoid as a
         single byte.
 
         :returns: The current value of all 8 solenoids on this module.
         """
-        return hal.getAllSolenoids(self.moduleNumber)
+        return SolenoidBase.getAll(self.moduleNumber)
```
auscompgeek: This self-recursive call will never work.

ariovistus: Whoops.

re "Static" - you answered yourself, its for backwards compatibility. For that particular naming convention, no reason. I'm not familiar enough with robotpy to know if a precedent has already been set for dealing with this particular situation. Figured I'd find out in the PR.

ariovistus:
you answered yourself, its for backwards compatibility. For that particular naming convention, no reason. I'm not familiar enough with robotpy to know if a precedent has already been set for dealing with this particular situation. Figured I'd find out in the PR.

auscompgeek:
If we're going to keep _both_ the static and instance methods, I'd rather the API isn't changed. We try to preserve upstream's API as much as we can in wpilib, so why stop now?

ariovistus:
Just making sure I'm not spacing here, the static methods are new, and the instance methods existed in previous years, ergo maintaining backwards compatibility would dictate the instance methods should not change names but the static methods can.

I suppose we could hack something together that differentiates between static and instance contexts and executes the right code. That way would get same name for both. Don't know how to do it off the top of my head, but python always seems to have a way

virtuald:
The time I really care about backwards compatibility is if we were the ones that created something. However, if upstream has created/removed a thing, then we should as well, backwards compatibility be damned -- unless there's a compelling reason to keep it.

ariovistus:
sounds to me like permission to hack granted

virtuald:
Hm, are you putting all of your various changes on the same PR? It would be easier to review them if they were in separate PRs... reviewing 43 files all at once is a bit difficult.

And, travis-ci isn't passing, so that needs to be resolved also (unless there's a compelling reason to ignore it), otherwise it makes it difficult to evaluate future PRs. :)

ariovistus:
due to tests, the changes are kind of interdependent

virtuald:
Ok.

ariovistus:
*cough* https://github.com/robotpy/pynetworktables/blob/master/networktables/entry.py#L449

virtuald:
Picky, picky...

virtuald:
Fine, one sec.

auscompgeek:
> To make the docs show it maybe put a third declaration of getAll above the static one?

I suspect that won't work. Sphinx doesn't do static analysis of Python code, AFAIK.

virtuald:
I pushed the NT entry changes, but it looks like travis is having capacity issues, so the build isn't starting.

ariovistus:
in the java code, listeners take a single parameter of type `EntryNotification`, are we going to do that too?

virtuald:
It doesn't really match the other listeners that we have... I think I prefer the explicit parameters instead of an opaque struct. The other problem is the underlying named tuple that gets passed in contains a `Value`, so you'd have to do `entry_notification.value.value`, or we would have to create a new named tuple and repack the data into that.

ariovistus:
there seems to be a null reference error in networktables, and something about hal not having time which I am unsure about

virtuald:
The networktables error is an ordering issue during reset... hm. That will require pyfrc changes I think.

The time thing is probably occurring because teardown is failing due to the prior error.

wpilib/wpilib/command/subsystem.py
```diff
@@ -59,7 +59,7 @@ def initDefaultCommand(self):
         """
         pass
 
-    def periodic():
```
virtuald: You should use an editor with a linter to avoid these sorts of errors. :)

I use atom, with the extra linter-pyflakes package installed.

wpilib/tests/test_pwm.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 from unittest.mock import MagicMock
+from ntcore.entry_notifier import _EntryNotification
```
virtuald: Why is this imported?

wpilib/tests/test_relay.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 from unittest.mock import MagicMock, call
+from ntcore.entry_notifier import _EntryNotification
```
virtuald: Why is this imported?

wpilib/tests/test_solenoid.py
```diff
@@ -1,5 +1,7 @@
 
 import pytest
+from unittest.mock import MagicMock
+from ntcore.entry_notifier import _EntryNotification
```
virtuald: Ditto.

wpilib/tests/test_solenoid.py
```diff
@@ -92,10 +103,12 @@ def test_solenoidbase_getAll(wpilib, hal_data):
         s['value'] = True
         
     assert solenoid.getAll() == 0xFF
+    assert wpilib.SolenoidBase.getAll(0) == 0xFF
```
virtuald: We should add a test for setAll()... but that means my hal changes need to be merged.

wpilib/wpilib/command/command.py
```diff
@@ -10,6 +10,7 @@
 from ..robotstate import RobotState
 from ..sendable import Sendable
 from ..timer import Timer
+from networktables.instance import NetworkTablesInstance
```
virtuald: I would rather you do `from networktables import NetworkTables`, and then use `NetworkTables.NotifyFlags`.. it reads better, and we will always have a global NetworkTables instance.

wpilib/wpilib/command/subsystem.py
```diff
@@ -140,14 +143,19 @@ def getSmartDashboardType(self):
     def initTable(self, table):
         super().initTable(table)
         if table is not None:
+            self.hasDefaultEntry = table.getEntry("hasDefault")
```
virtuald: These four should probably be initialized to `None` in the constructor.

wpilib/wpilib/pidcontroller.py
```diff
@@ -10,6 +10,7 @@
 from itertools import islice
 import threading
 import warnings
+from networktables.instance import NetworkTablesInstance
```
virtuald: Ditto.

wpilib/wpilib/pidcontroller.py
```diff
@@ -519,41 +507,91 @@ def reset(self):
...
+            self.dEntry = table.getEntry("d")
+            self.dEntry.setDouble(self.getD())
+            self.fEntry = table.getEntry("f")
+            self.fEntry.setDouble(self.getF())
+            self.setpointEntry = table.getEntry("setpoint")
+            self.setpointEntry.setDouble(self.getSetpoint())
+            self.enabledEntry = table.getEntry("enabled")
+            self.enabledEntry.setBoolean(self.isEnabled())
+
+            self.pListener = self.pEntry.addListener(
```
virtuald: Ouch. Adding all of these listeners is going to end up with a ton of overhead. I know you didn't do this so no fix required here -- but we need to file a bug upstream.

virtuald: Well.. on second thought, if that's only active during LiveWindow mode, then I don't really care. Unless it's active all the time?

wpilib/tests/test_relay.py
ariovistus: oops, it was used until tonight

virtuald:
The NT entry error shouldn't be happening... the only way this can happen is if the entry was created before NT.shutdown() was called and then the removal happens after shutdown occurs. In pyfrc the teardown first does a reset in wpilib, and then shuts down networktables... so all of the entries should go away. Each entry holds a pointer to the NtCoreApi object, which gets all of its internals cleaned out at shutdown.

This could be an instance of https://github.com/robotpy/examples/issues/5, but gearsbot doesn't use globals. There must be some global state that isn't being reset somewhere, but I don't see where it could be.

wpilib/tests/test_solenoid.py
virtuald: My hal changes are merged, so someone should add a setAll test...

virtuald:
I can't debug the global issue tonight, but maybe tomorrow. It's weird.

wpilib/tests/test_solenoid.py
ariovistus: where is setAll?

virtuald: They just introduced something at the HAL layer, so I had made an assumption that it would show up in the high level API too:

https://github.com/wpilibsuite/allwpilib/blob/6e4f66cc8b9fd96d82ca4119fc337bfc00331f09/hal/src/main/native/athena/Solenoid.cpp#L124

However, my assumption is incorrect and you're right. It does not appear to be there.

ariovistus:
just checking in, what's the status of this one? I don't even know where the tests are that are failing or how I would reproduce locally

Twinters007:
```
/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/pyfrc/test_support/pytest_plugin.py:96: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../wpilib/wpilib/_impl/utils.py:205: in reset_wpilib
    cls._reset()
../../wpilib/wpilib/resource.py:51: in _reset
    obj.free()
../../wpilib/wpilib/pidcontroller.py:151: in free
    self.removeListeners()
../../wpilib/wpilib/pidcontroller.py:512: in removeListeners
    self.pEntry.removeListener(self.pListener)
/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/networktables/entry.py:471: in removeListener
    self.__api.removeEntryListener(listener_id)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <ntcore.api.NtCoreApi object at 0x7f3347db3b38>, listener_uid = 8
    def removeEntryListener(self, listener_uid):
>       self.entry_notifier.remove(listener_uid)
E       AttributeError: 'NoneType' object has no attribute 'remove'
```

virtuald:
This PR is too big to really digest easily, so I haven't been able to look into it yet as I don't have the cycles. If you can extract some of the independent pieces that don't break the tests off into individual PRs, I'd love to merge those, and it would probably make it easier to figure out what is wrong with the tests for this one.

ariovistus:
There are a few trivial files that could be split out, but I think the big problem is Sendable changed and you need it changed if a subclass is changed, and then to preserve tests passing you need all subclasses changed. I guess you could provide a TempNewSendable for the sake of smaller PRs

PR #286, opened 2017-10-07
update motor controllers for 2018

ariovistus:


virtuald:
ntcore is fixed, please rebase. Thanks!

PR #292, opened 2017-10-23
docs: Skip undocumented classes

auscompgeek:
This causes the docs to skip generating documentation for classes that are not
documented.  This is intended to skip the CANTalon and CANJaguar stubs that were
added in 2017 to direct users to the relevant packages (or lack thereof).

This also documents a couple of classes in the buttons and interfaces packages
that were meant to be documented, but weren't.

I have verified that there are no other classes that are undocumented using the
following snippet:
```python
[x for package in [wpilib, wpilib.buttons, wpilib.command, wpilib.interfaces]
   for x in vars(package).values() if inspect.isclass(x) and not x.__doc__]
```

Fixes: #268

### DIFF
--- a/docs/regen.py
+++ b/docs/regen.py
@@ -80,6 +80,9 @@ def gen_package(mod):
     classes = []
     
     for clsname, cls in inspect.getmembers(mod, inspect.isclass):
+        # Skip undocumented classes - only stubs will be undocumented.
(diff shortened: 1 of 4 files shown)


auscompgeek:
Hmm, maybe we should also add a test checking that every class is documented.

Issue #38, opened 2014-11-21
PIDController does not have deterministic timing


PeterJohnson:
As currently written, PIDController doesn't repeat at deterministic intervals because it uses a timed sleep instead of being started periodically.  For example, if the PID get or set functions take variable amounts of time to run, the loop timing changes, which is undesirable for controller time-dependent I and D gain settings.

virtuald:
From my experiments with PreciseDelay, I suspect that this fix was better than nothing, but isn't a complete fix. Something similar to [PreciseDelay](https://github.com/robotpy/robotpy-wpilib-utilities/blob/master/robotpy_ext/misc/precise_delay.py) should be implemented.

ThadHouse:
One thing I noticed is that the C++ code uses the FPGA Notifier to run the PIDController. I know that 254 submitted a patch that added Notifier support to Java. If Notifier support was added here, it could potentially be used for PIDController to give it better timing. 

254s patch does have a bug in it where it gets stuck in an infinite while loop with more then 2 notifiers, but its fixable with 2 lines of code.

I don't know enough python to implement this myself, but maybe over the summer I'll learn enough to try and do it.

virtuald:
@ariovistus ... so in your testing, is this bug moot then?

ariovistus:
for period in the neighborhood of 1 ms, notifier based is definitely better.

virtuald:
Yeah... but if you're using 1ms notifier you're probably going to have other problems.

ariovistus:
like what? (is late, my mind isn't moving quickly)

auscompgeek:
You'd probably have GIL contention problems at that rate.

PR #297, opened 2017-11-09
Add TimedRobot

twinters007:


virtuald:
You know you can run the unit tests locally, right? ;)

Twinters007:
I like to live dangerously

auscompgeek:
I don't think IterativeRobotBase belongs in `interfaces/`; it's not an interface, and RobotBase doesn't live there either.

wpilib/wpilib/timedrobot.py
```diff
@@ -0,0 +1,33 @@
...
+from interfaces.iterativerobotbase import IterativeRobotBase
+from robotpy_ext.misc import PreciseDelay
+
+
+class TimedRobot(IterativeRobotBase):
+    DEFAULT_PERIOD = .02
+    period = DEFAULT_PERIOD
+
+    def __init__(self):
+        hal.report(hal.UsageReporting.kResourceType_Framework, hal.UsageReporting.kFramework_Iterative)
```
auscompgeek: Missing `super().__init__()` call.

Twinters007:
Oh yeah I didn't realize wpilib had an interfaces class. I thought abstract classes might go there.

virtuald:
Naturally, there are conflicts.

auscompgeek:
Given that we still haven't quite yet come to a consensus on how we should implement TimedRobot, and the LiveWindow/Sendable changes upstream touch almost everything (including IterativeRobotBase), shall we deal with the IterativeRobot changes now and table TimedRobot for later?

Users can always implement their own TimedRobot (perhaps the same way as this PR) in the interim, if they prefer.

auscompgeek:
Oh, #38 has similar discussion.

Twinters007:
I think (especially since Dustin is unsure about the future of RobotPy) we should move PreciseDelay to robotpy-wpilib as a notifier alternative, use it in PIDController and TimedRobot, and then, if needed later, use a notifier

virtuald:
PreciseDelay is not a good notifier alternative. It's good for delay of a single thread, not many threads.

auscompgeek:
Yeah, I don't think PreciseDelay will behave well in a multithreaded environment.

wpilib/wpilib/iterativerobotbase.py
```diff
@@ -0,0 +1,226 @@
...
+
+    last_mode = Mode.kNone
+
+    #: A python logging object that you can use to send messages to the log. It
+    #: is recommended to use this instead of print statements.
+    logger = logging.getLogger("robot")
+
+    def __init__(self):
+        super().__init__()
+        hal.observeUserProgramStarting()
```
auscompgeek: Uhm. I think this got out of sync with upstream. This doesn't match the commit in the git-source-track line.

wpilib/wpilib/iterativerobotbase.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def startCompetition(self):
+        raise NotImplementedError("You haven't implemented startCompetition")
```
auscompgeek: This isn't needed here, right?

Twinters007: Yes it is?

auscompgeek: Why? Just let it go through the inheritance chain.

Twinters007: Oh you mean let RobotBase give the error?

auscompgeek: Hm. I think it's there for documentation sake. This being here has _negative_ utility if it's not documented.

Twinters007: Yeah idk why they implemented it just to keep it abstract...If I get rid of it here, I wanna move the message because `RobotBase` doesn't give anything

auscompgeek: The message is tautological, realistically. It's in the traceback.

Twinters007: Eh true

wpilib/wpilib/iterativerobotbase.py
auscompgeek: This still very much does not match, and I think there's a good reason why it _isn't_ here upstream.

wpilib/wpilib/iterativerobotbase.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+
+    class Mode:
+        kNone = 0
+        kDisabled = 1
+        kAutonomous = 2
+        kTeleop = 3
+        kTest = 4
+
+    last_mode = Mode.kNone
```
auscompgeek: Set this in `__init__` instead? Users probably don't care about this implementation detail.

Twinters007: Why does putting it in init help?

wpilib/wpilib/iterativerobotbase.py
```diff
@@ -0,0 +1,222 @@
+# validated: 2017-11-09 TW ef3267833fc3 edu/wpi/first/wpilibj/IterativeRobotBase.java
```
auscompgeek: Update IterativeRobotBase for the LiveWindow changes?

wpilib/wpilib/iterativerobotbase.py
```diff
@@ -0,0 +1,222 @@
...
+
+    periodic() functions -- each of these functions is called on an interval:
+    - robotPeriodic()
+    - disabledPeriodic()
+    - autonomousPeriodic()
+    - teleopPeriodic()
+    - testPeriodic()
+    """
+
+    class Mode:
```
auscompgeek: Make this an enum.Enum?

wpilib/wpilib/iterativerobotbase.py
```diff
@@ -0,0 +1,222 @@
...
+        kNone = 0
+        kDisabled = 1
+        kAutonomous = 2
+        kTeleop = 3
+        kTest = 4
+
+
```
auscompgeek: *cough*

wpilib/wpilib/timedrobot.py
```diff
@@ -0,0 +1,34 @@
+# validated: 2017-11-09 TW ef3267833fc3 edu/wpi/first/wpilibj/TimedRobot.java
```
auscompgeek: I'd like to pull in the IterativeRobotBase changes ASAP. I say TimedRobot can wait for another PR.

Twinters007: It'll have to wait anyway for the notifier

auscompgeek: Yeah, but this PR is now holding up actually finishing the LiveWindow changes.

Twinters007: Yeah I agree. Now I gotta figure out the best way to separate IterativeRobotBase...

Twinters007:
I'm probably gonna remove iterative, since I have all of it in #381 anyway. Where is @ariovistus when we need him

ariovistus:
sleeping. what's up?

Twinters007:
Now relies on #381

.gittrack
```diff
@@ -4,3 +4,4 @@ exclude_commits_file = devtools/exclude_commits
 upstream_root = ../allwpilib/wpilibj/src/main/java
 upstream_commit = 166d9e01bf756b1e72309e5a0de9dc0177d26e81
 
+
```
auscompgeek: Not really sure what you did here.

wpilib/wpilib/timedrobot.py
```diff
@@ -0,0 +1,52 @@
...
+    def startCompetition(self):
+        self.robotInit()
+
+        hal.observeUserProgramStarting()
+
+        self.startLoop = True
+        self.loop.startPeriodic(self.period)
+
+        while True:
+            Timer.delay(60*60*24)
```
auscompgeek: Do things break if this were a `time.sleep()` call instead?

.gittrack
twinters007: Just trying to mess with you at this point

wpilib/wpilib/timedrobot.py
twinters007: No but I figured it better to use hal.sleep for...reasons?

ariovistus: simulation?

twinters007: What do you mean?

wpilib/wpilib/timedrobot.py
```diff
@@ -0,0 +1,52 @@
...
+from .iterativerobotbase import IterativeRobotBase
+from .notifier import Notifier
+from .timer import Timer
+
+
+__all__ = ["TimedRobot"]
+
+
+class TimedRobot(IterativeRobotBase):
+    DEFAULT_PERIOD = .02
```
auscompgeek: Where's the documentation?

wpilib/wpilib/timedrobot.py
```diff
@@ -0,0 +1,52 @@
...
+
+        while True:
+            Timer.delay(60*60*24)
+
+
+    def setPeriod(self, period):
+        """Set time period between calls to Periodic() functions.
+
+        :param period: Period in seconds
+        :type period: float
```
auscompgeek: Should use type hints :)

twinters007: tru

wpilib/wpilib/timedrobot.py
twinters007: 🤷🏼‍♀️

wpilib/wpilib/timedrobot.py
virtuald: We should add it to the porting notes -- never use time.sleep, because simulation reasons.

twinters007: Wow I did something right for once

ariovistus: what is this sleep doing, anyways? it looks like the intent is to keep the calling thread alive but not consuming cpu resources. would hal sleep vs system sleep really make any difference?

auscompgeek: I suspect tests would hang if you called `time.sleep()`. It'd be fine on the real robot though.

PR #302, opened 2017-11-16
update solenoid and solenoidbase for 2018

ariovistus:


wpilib/wpilib/solenoidbase.py
```diff
@@ -22,14 +22,30 @@ def __init__(self, moduleNumber):
         :param moduleNumber: The PCM CAN ID
         """
         self.moduleNumber = moduleNumber
+        self.getAll = lambda: SolenoidBase.getAll(self.moduleNumber)
```
virtuald: I know it was discussed previously... but I've already forgotten, what's this weirdness for again? Since there are two functions called `getAll`...

virtuald:
Thanks for splitting this up, it's a lot easier to determine if the changes are mergeable.

wpilib/wpilib/solenoidbase.py
Twinters007: Also, wouldn't it be better to just do `self.getAll = SolenoidBase.getAll` (provided this is the feature you wanted

wpilib/tests/test_solenoid.py
```diff
@@ -92,10 +102,12 @@ def test_solenoidbase_getAll(wpilib, hal_data):
         s['value'] = True
         
     assert solenoid.getAll() == 0xFF
+    assert wpilib.SolenoidBase.getAll(0) == 0xFF
```
ariovistus: the getAll weirdness: upstream they added a static getAll with a parameter, but left the old instance method with no parameter, so to get the same behavior in python requires some hacking

wpilib/wpilib/solenoidbase.py
ariovistus: the first declared getAll (line 32) is my feeble attempt to get the docs to show that you can call getAll from both static and nonstatic contexts with the different signatures. It probably won't work and if not should just be deleted

wpilib/tests/test_solenoid.py
```diff
@@ -79,6 +80,15 @@ def test_multiple_solenoids(wpilib, hal_data):
             assert s.get() == nv
     
 
+@pytest.mark.parametrize("value,expected",
+    [(True, True), (False, False)])
```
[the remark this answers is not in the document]
ariovistus: fine, I'll make them not be the same

wpilib/wpilib/solenoidbase.py
auscompgeek: Given I'm pretty sure Sphinx imports the code and introspects what's available at runtime, rather than doing static analysis, this almost certainly won't work.

auscompgeek:
Instead of using `@staticmethod`, why not have an `isinstance()` check in a regular method, or use `@functools.singledispatch`?

Issue #239, opened 2017-01-06
Remove calls to MagicMock in tests


james-ward:
Use the HAL mocking mechanism instead.

virtuald:
OBE.

PR #304, opened 2017-11-16
update relay for 2018

ariovistus:


wpilib/wpilib/relay.py
```diff
@@ -258,26 +278,17 @@ def setDirection(self, direction):
     def getSmartDashboardType(self):
         return "Relay"
 
+    def initTable(self, subtable):
+        if subtable is not None:
+            self.valueEntry = subtable.getEntry("Value")
+            self.updatTable()
```
virtuald: updatTable?

auscompgeek:
So this also fixes #253. It would be nice if we could get tests for that too, and maybe fix the HAL as well.

virtuald:
We weren't using `enum.IntEnum` and friends because we were discussing in 2015 whether to support Python 2.7 or not... and I think it was introduced in 3.4, so 3.3 support was vaguely important for a small period of time.

I think we could switch everything to use enums now -- but I think that would be a separate (massive?) PR.

virtuald:
Also, we really need to get rid of all the Mock tests, as they don't do any typechecking when it decends down into HAL, see #239

PR #313, opened 2017-11-16
update command/scheduler.py for 2018

ariovistus:


wpilib/wpilib/command/subsystem.py
```diff
@@ -59,6 +64,11 @@ def initDefaultCommand(self):
         """
         pass
 
+    def periodic(self):
+        """When the run method of the scheduler is called this method will be called.
+        """
+        pass
```
Twinters007: I imagine you want them to overload this? You should print some message saying you want to be overloaded. 
Something like 
```
func = self.periodic.__func__
if not hasattr(func, "firstRun"):
self.logger.info("Default IterativeRobot.robotPeriodic() method... Overload me!")
func.firstRun = False

ariovistus: that does appear to be the intent

PR #316, opened 2017-11-17
update pidcontroller for 2018

ariovistus:


wpilib/wpilib/pidcontroller.py
```diff
@@ -519,41 +507,91 @@ def reset(self):
             self.totalError = 0
             self.result = 0
 
+    def removeListeners(self):
+        if self.pEntry is not None:
+            self.pEntry.removeListener(self.pListener)
```
virtuald: The `None` exception is occurring because you aren't setting `self.pEntry` to None (though I'm not 100% sure why this is).

Twinters007: > if it's not none
> code fails because it's none
🤔

virtuald: I figured it out, the global resources weren't getting cleared out, so free/reset were being called multiple times because the object never went away.

However, it's probably a good idea to set all the entries/etc to None when removing them, so let's do that.

PR #317, opened 2017-11-19
update things that don't need to be updated for 2018

ariovistus:


ariovistus:
hang on here, somebody updated upstream

Twinters007:
Can you make a better commit message?

ariovistus:
what are those stamps called? source-track stamps?

Twinters007:
You could just say `verified non-changing files`

virtuald:
One of your other PRs must have created a conflict.

PR #321, opened 2017-11-21
Fix logic in vector2d.

Twinters007:


auscompgeek:
You forgot to fix the test as well.

Twinters007:
What's wrong with the tests? There's no variable assignment

auscompgeek:
The test is literally broken.

That is, you're asserting that `vector.x`.

Twinters007:
Ah.

PR #323, opened 2017-11-22
update compressor.py for 2018

ariovistus:


wpilib/wpilib/compressor.py
```diff
@@ -148,14 +152,34 @@ def getSmartDashboardType(self):
...
-        table = self.getTable()
-        if table is not None:
-            table.putBoolean("Enabled", self.enabled())
-            table.putBoolean("Pressure Switch", self.getPressureSwitchValue())
+        if self.enabledEntry is not None:
+            self.enabledEntry.setBoolean(self.enabled())
+        if self.pressureSwitchEntry is not None:
+            self.pressureSwitchEntry.setBoolean(self.getPressureSwitchValue())
+
+    def startLiveWindowMode(self):
```
virtuald: Hm? Isn't this done in liveWindowSendable?

ariovistus: what? setting up the listener? wrong entry

ariovistus:
added a bunch of tests (and found some bugs). hal is breaking on test_compressor_clearallpcmstickyfaults, but it's kind of opaque, not sure if it's a problem with Compressor, tests, or hal

wpilib/wpilib/compressor.py
auscompgeek: Well, only one entry is on the LiveWindow here. Simple fix.

ariovistus: ?

virtuald:
`clearAllPCMStickyFaults` was implemented incorrectly in the simulated HAL. Examining the declaration functions.py shows that it has no return value, and the simulated function was returning `False`.

PR #331, opened 2017-11-23
Update digital files

Twinters007:
digitalinput
digitaloutput
interruptablesensorbase
digitalsource

### DIFF
--- a/wpilib/wpilib/digitalinput.py
+++ b/wpilib/wpilib/digitalinput.py
@@ -1,10 +1,10 @@
-# validated: 2017-10-24 EN 34c18ef00062 edu/wpi/first/wpilibj/DigitalInput.java
-#----------------------------------------------------------------------------
+# validated: 2017-11-23 TW 34c18ef00062 edu/wpi/first/wpilibj/DigitalInput.java
+# ----------------------------------------------------------------------------
 # Copyright (c) FIRST 2008-2012. All Rights Reserved.
 # Open Source Software - may be modified and shared by FRC teams. The code
 # must be accompanied by the FIRST BSD license file in the root directory of
 # the project.
-#----------------------------------------------------------------------------
+# ----------------------------------------------------------------------------
 
 import hal
 
@@ -12,6 +12,7 @@ from .digitalsource import DigitalSource
 from .livewindow import LiveWindow
 
 __all__ = ["DigitalInput"]
+
 
 class DigitalInput(DigitalSource):
     """Reads a digital input.
@@ -30,13 +31,16 @@ class DigitalInput(DigitalSource):
         :param channel: the DIO channel for the digital input. 0-9 are on-board, 10-25 are on the MXP
         :type  channel: int
         """
-        # Store the channel and generate the handle in the constructor of the parent class
-        # This is different from the Java implementation
-        super().__init__(channel, True)
 
-        hal.report(hal.UsageReporting.kResourceType_DigitalInput,
-                      channel)
+        super().__init__()
+        self.checkDigitalChannel(channel)
+        self.channel = channel
+
+        self.handle = hal.initializeDIOPort(hal.getPort(channel), True)
+
         LiveWindow.addSensor("DigitalInput", channel, self)
+        hal.report(hal.UsageReporting.kResourceType_DigitalInput,
+                   channel)
 
         self.valueEntry = None
 
@@ -44,7 +48,7 @@ class DigitalInput(DigitalSource):
         if self.interrupt:
             self.cancelInterrupts()
 
-        super().free() # This calls hal.freeDIOPort
+        hal.freeDIOPort(self.handle)
 
     def get(self):
         """Get the value from a digital input channel. Retrieve the value of
@@ -106,4 +110,3 @@ class DigitalInput(DigitalSource):
 
     def stopLiveWindowMode(self):
         pass
-
--- a/wpilib/wpilib/digitaloutput.py
+++ b/wpilib/wpilib/digitaloutput.py
@@ -1,10 +1,10 @@
-# validated: 2017-11-19 EN b65447b6f5a8 edu/wpi/first/wpilibj/DigitalOutput.java
-#----------------------------------------------------------------------------
+# validated: 2017-11-23 TW b65447b6f5a8 edu/wpi/first/wpilibj/DigitalOutput.java
+# ----------------------------------------------------------------------------
 # Copyright (c) FIRST 2008-2012. All Rights Reserved.
 # Open Source Software - may be modified and shared by FRC teams. The code
 # must be accompanied by the FIRST BSD license file in the root directory of
 # the project.
-#----------------------------------------------------------------------------
+# ----------------------------------------------------------------------------
 
 import warnings
 import weakref
@@ -28,14 +28,21 @@ class DigitalOutput(DigitalSource):
     digital inputs and outputs as required.
     """
 
+    invalidPwmGenerator = None
+
     def __init__(self, channel):
         """Create an instance of a digital output.
 
         :param channel: the DIO channel for the digital output. 0-9 are on-board, 10-25 are on the MXP
         """
-        super().__init__(channel, False)
+
         self._pwmGenerator = None
         self._pwmGenerator_finalizer = None
+
+        self.checkDigitalChannel(channel)
+        self.channel = channel
+
+        self.handle = hal.initializeDIOPort(hal.getPort(channel), False)
 
         hal.report(hal.UsageReporting.kResourceType_DigitalOutput,
                       channel)
@@ -56,8 +63,10 @@ class DigitalOutput(DigitalSource):
         if self.pwmGenerator is not None:
             self._pwmGenerator_finalizer()
         self._pwmGenerator = None
-        super().free()
-        self._handle = None
+        if self.pwmGenerator is not self.invalidPwmGenerator:
+            self.disablePWM()
+        hal.freeDIOPort(self.handle)
+        self.handle = 0
 
     def set(self, value):
         """Set the value of a digital output.
@@ -65,7 +74,7 @@ class DigitalOutput(DigitalSource):
         :param value: True is on, off is False
         :type  value: bool
         """
-        hal.setDIO(self.handle, value)
+        hal.setDIO(self.handle, bool(value))
 
     def get(self):
         """Gets the value being output from the Digital Output.
@@ -124,7 +133,7 @@ class DigitalOutput(DigitalSource):
         :param initialDutyCycle: The duty-cycle to start generating. [0..1]
         :type  initialDutyCycle: float
         """
-        if self.pwmGenerator is not None:
+        if self.pwmGenerator is not self.invalidPwmGenerator:
             return
         self._pwmGenerator = hal.allocateDigitalPWM()
         hal.setDigitalPWMDutyCycle(self._pwmGenerator, initialDutyCycle)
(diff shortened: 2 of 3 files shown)


wpilib/wpilib/digitalinput.py
```diff
@@ -32,19 +33,23 @@ def __init__(self, channel):
         """
         # Store the channel and generate the handle in the constructor of the parent class
         # This is different from the Java implementation
-        super().__init__(channel, True)
+        super().__init__()
```
auscompgeek: You realise the comment no longer matches, right?

Twinters007: Nope

wpilib/wpilib/interruptablesensorbase.py
```diff
@@ -97,28 +90,33 @@ def waitForInterrupt(self, timeout, ignorePrevious=True):
...
         rising = 0x1 if (result & 0xFF) else 0x0
         falling = 0x100 if (result & 0xFF00) else 0x0
 
-        return rising | falling
+        result = rising | falling
+        for mode in self.WaitResult:
+            if mode == result:
+                return mode
+
+        return None
```
auscompgeek: Seeing as you've made WaitResult an enum, use `self.WaitResult(result)` instead.

wpilib/wpilib/sensorbase.py
```diff
@@ -1,18 +1,19 @@
...
 
 import hal
 
 from .livewindowsendable import LiveWindowSendable
 
 __all__ = ["SensorBase"]
 
-class SensorBase(LiveWindowSendable):
+
+class SensorBase:
```
auscompgeek: If you're going to remove this, you need to make everything that should be a LiveWindowSendable inherit from LiveWindowSendable…

wpilib/wpilib/interruptablesensorbase.py
```diff
@@ -1,26 +1,33 @@
...
 
+
 class InterruptableSensorBase(SensorBase):
     """Base for sensors to be used with interrupts"""
 
+    class WaitResult(enum.IntEnum):
+        kTimeout = 0x0
+        kRisingEdge = 0x1
+        kFallingEdge = 0x100
+        kBoth = 0x101
```
auscompgeek: Shame `enum.IntFlag` was only introduced in Python 3.6. It'd be a really great fit here…

wpilib/wpilib/sensorbase.py
auscompgeek: I think there are a lot more sensors in WPILib that you haven't touched here that should be LiveWindowSendable.

wpilib/wpilib/digitaloutput.py
```diff
@@ -124,23 +119,25 @@ def enablePWM(self, initialDutyCycle):
         :param initialDutyCycle: The duty-cycle to start generating. [0..1]
         :type  initialDutyCycle: float
         """
-        if self.pwmGenerator is not None:
+        if self.pwmGenerator is not self.invalidPwmGenerator:
```
virtuald: I think you mean `is`

wpilib/wpilib/interruptablesensorbase.py
```diff
@@ -70,22 +67,18 @@ def allocateInterrupts(self, watcher):
...
-        self._interrupt_finalizer = weakref.finalize(self, hal.cleanInterrupts,
-                                                     self._interrupt)
+        self.interrupt = hal.initializeInterrupts(watcher)
 
     def cancelInterrupts(self):
         """Cancel interrupts on this device. This deallocates all the
         chipobject structures and disables any interrupts.
         """
-        if self.interrupt is None:
+        if self.interrupt is 0:
```
virtuald: `if not self.interrupt`

wpilib/wpilib/sensorbase.py
```diff
@@ -68,7 +69,8 @@ def checkSolenoidModule(moduleNumber):
         :param moduleNumber: The solenoid module module number to check.
         """
         if not hal.checkSolenoidModule(moduleNumber):
-            raise IndexError("Requested solenoid module number %d is out of range [0, %d)." % (moduleNumber, SensorBase.kPCMModules))
```
virtuald: Cmon, if you're going to do a formatting PR, do it separately. This makes reviewing really annoying.

wpilib/wpilib/interruptablesensorbase.py
```diff
@@ -1,26 +1,33 @@
...
+        kRisingEdge = 0x1
+        kFallingEdge = 0x100
+        kBoth = 0x101
+
     def __init__(self):
         """Create a new InterrupatableSensorBase"""
         # The interrupt resource
-        self._interrupt = None
-        self._interrupt_finalizer = None
+        self.interrupt = 0
```
virtuald: Why are we changing this. To match? Semantically, `None` is a more pythonic representation of absence.

wpilib/wpilib/sensorbase.py
virtuald: Yeah... we should keep this I think? And document that it's python-specific.

wpilib/wpilib/sensorbase.py
auscompgeek: Not to mention that the reformatting feels really awkward here.

wpilib/wpilib/sensorbase.py
Twinters007: Is that the final verdict?

wpilib/wpilib/digitaloutput.py
virtuald: Nevermind, I can't read.

wpilib/wpilib/sensorbase.py
ariovistus: A bunch of sensors depend on it to set up their value listeners, and this behavior isn't covered by unit tests for most of them. In the interest of getting things done, it should be kept, though personally, I would rather refactor it. As is, it isn't easy to see when this is happening, and it only works for "Value" entries; if it were an explicit call it could be more generic

hal-sim/hal_impl/functions.py
```diff
@@ -856,7 +856,24 @@ def freeDIOPort(dioPortHandle):
 
 def allocateDigitalPWM(status):
```
Twinters007: @virtuald / @auscompgeek can you confirm that this is correct logically and the correct way to go about this? I think I interpreted the C code correctly, but not 100% sure.

hal-sim/hal_impl/functions.py
```diff
@@ -856,7 +856,24 @@ def freeDIOPort(dioPortHandle):
...
+                break
+        if not port['initialized']:
+            dio = _initport('dio', i, status)
+            handle = i
+            break
+
+    if dio is None:
+        return 0
+
+    return types.DigitalPWMHandle(getPort(handle))
```
Twinters007: @virtuald / @auscompgeek can you confirm that this is correct logically and the correct way to go about this? I think I interpreted the C code correctly, but not 100% sure.

auscompgeek: I'm having trouble figuring out how you got to this implementation. `HAL_AllocateDigitalPWM` is _very_ different…

wpilib/wpilib/interruptablesensorbase.py
```diff
@@ -1,26 +1,33 @@
...
 from .sensorbase import SensorBase
 
 __all__ = ["InterruptableSensorBase"]
 
+
 class InterruptableSensorBase(SensorBase):
     """Base for sensors to be used with interrupts"""
 
+    class WaitResult(enum.IntEnum):
+        kTimeout = 0x0
```
virtuald: Style thought, I would probably pad each of these out to 4 digits... `0x0000`, `0x0100`, etc.

virtuald:
There are conflicts. Also, can you break this into multiple PRs? Probably one for interrupt, the other for the weird PWM stuff, and the other for the other two?

I'm not sure I understand what's going on with your changes to the PWM stuff.

wpilib/wpilib/digitaloutput.py
```diff
@@ -1,71 +1,65 @@
...
 
+
 class DigitalOutput(DigitalSource):
     """Writes to a digital output
     
     Other devices that are implemented elsewhere will automatically allocate
     digital inputs and outputs as required.
     """
 
+    invalidPwmGenerator = 0
```
auscompgeek: Note that this is a private constant upstream. I think we are perfectly fine with using `None` for this instead.

wpilib/wpilib/digitaloutput.py
```diff
@@ -1,71 +1,65 @@
...
 from .digitalsource import DigitalSource
 from .sensorbase import SensorBase
 
 __all__ = ["DigitalOutput"]
 
+
 def _freePWMGenerator(pwmGenerator):
     # Disable the output by routing to a dead bit.
     hal.setDigitalPWMOutputChannel(pwmGenerator, SensorBase.kDigitalChannels)
     hal.freeDigitalPWM(pwmGenerator)
```
auscompgeek: With your changes, this function is no longer used. Either use this or delete it.

wpilib/wpilib/digitaloutput.py
```diff
@@ -124,23 +118,25 @@ def enablePWM(self, initialDutyCycle):
...
             return
-        self._pwmGenerator = hal.allocateDigitalPWM()
-        hal.setDigitalPWMDutyCycle(self._pwmGenerator, initialDutyCycle)
-        hal.setDigitalPWMOutputChannel(self._pwmGenerator, self.channel)
-        self._pwmGenerator_finalizer = \
-                weakref.finalize(self, _freePWMGenerator, self._pwmGenerator)
+
+        self.pwmGenerator = hal.allocateDigitalPWM()
+        hal.setDigitalPWMDutyCycle(self.pwmGenerator, initialDutyCycle)
+        hal.setDigitalPWMOutputChannel(self.pwmGenerator, self.channel)
```
ariovistus: any reason for removing the finalizer? it doesn't look like enablePWM has changed, and one would assume that the original reason for having it still holds. It looks to me like the finalizer ensures the pwm channel gets deallocated, even if disablePWM isn't called.

wpilib/wpilib/digitalsource.py
```diff
@@ -15,55 +15,22 @@
 
 __all__ = ["DigitalSource"]
 
+
 def _freeDigitalSource(handle):
     hal.freeDIOPort(handle)
```
ariovistus: this one looks to be unused now

wpilib/wpilib/digitaloutput.py
Twinters007: https://github.com/robotpy/robotpy-wpilib/pull/346

Take a look at this comment.

Twinters007: Because I don't understand complex python techniques

Twinters007: Mostly it boils down to me not knowing advanced python techniques

ariovistus: ugh, I had even read that comment. It sounds to me like the finalizer should stay in. If the FPGA actually does recycle things for you, then maybe you don't actually need it, but it doesn't stop being a good practice.

Twinters007: Ok I'll put it back in, and let a more knowledgeable person remove it

virtuald:
Bring the finalizers back and we can merge this.

auscompgeek:
This doesn't look so bad now, although I'm currently reviewing this on my phone.

Spelling out `invalidPwmGenerator` seems a bit verbose though when it's just None.

Twinters007:
Hmmm. Makes it clearer?

@virtuald I think I got all the finalizers back

wpilib/wpilib/digitaloutput.py
```diff
@@ -138,8 +147,10 @@ def disablePWM(self):
 
         Free up one of the 6 DO PWM generator resources that were in use.
         """
-        if self.pwmGenerator is None:
+        if self.pwmGenerator is not self.invalidPwmGenerator:
             return
+        hal.setDigitalPWMOutputChannel(self.pwmGenerator, self.kDigitalChannels)
```
Twinters007: I guess I shouldn't be swapping between `self._pwmGenerator` and `self.pwmGenerator`

PR #334, opened 2017-11-29
add ultrasonic unit tests

ariovistus:


auscompgeek:
_coughs_

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called

ariovistus:
well that's inconvenient

wpilib/tests/test_ultrasonic.py
```diff
@@ -1,27 +1,155 @@
...
+def test_ultrasonic_updateTable(ultrasonic, hal_data):
+    ultrasonic.ping()
+    hal_data['counter'][0]['count'] = 2
+    hal_data['counter'][0]['period'] = 0.0012
+
+    ultrasonic.valueEntry = MagicMock()
+
+    ultrasonic.updateTable()
+
+    ultrasonic.valueEntry.setDouble.assert_called_with(pytest.approx(8.13, 0.01))
```
auscompgeek: Instead of mocking a value entry, why not check that NetworkTables is updated correctly? The tests should check the expected behaviour, not the implementation.

wpilib/tests/test_ultrasonic.py
```diff
@@ -1,27 +1,155 @@
...
+
+    assert not hasattr(ultrasonic, 'valueListener')
+
+
+def test_stoplivewindowmode_null_entry(ultrasonic):
+    ultrasonic.valueEntry = None
+
+    ultrasonic.startLiveWindowMode()
+
+    ultrasonic.stopLiveWindowMode()
```
auscompgeek: This looks like it belongs in tests for `LiveWindowSendable` instead.

wpilib/tests/test_ultrasonic.py
ariovistus: oh hey, networktables is set up as a fixture

wpilib/tests/test_ultrasonic.py
ariovistus: that assumes Ultrasonic subclasses LiveWindowSendable, which is an implementation detail that might change in the future

virtuald: I think it belongs here.

auscompgeek: I'm still slightly sceptical, because there shouldn't be a scenario where the `startLiveWindowMode()` method on an Ultrasonic object should be called before `initTable()`.

PR #335, opened 2017-11-29
add more unit tests for Solenoid and DoubleSolenoid

ariovistus:


ariovistus:
couple issues: Solenoid is missing `setPulseDuration` and `startPulse`, and hal seems to be missing supporting functions
Also, in `test_doublesolenoid_livewindowmode`, I'm testing listener behavior and it isn't doing calling the callback, does something need to be kicked to make this work in tests, is this inadvisable, or is something else wrong?

auscompgeek:
Where are you seeing `Solenoid.setPulseDuration()` and `Solenoid.startPulse()`? They're definitely not in the tree at wpilibsuite/allwpilib@303c259b89fee57cbdb067bc18712beccf4e26a4.

ariovistus:
looks like it was just added: https://github.com/wpilibsuite/allwpilib/commit/7a250a1b933e5f53baca6ac03fe08629d0f12bb3#diff-2074bbc44fb471588868a90710dda71f

virtuald:
I think LiveWindow stuff only gets activated in Test mode?

virtuald:
Also, working on updating HAL right now, will push soon.

ariovistus:
So from my read of pynetworktables earlier today, it looks like there is a separate thread, EntryNotifierThread, that drives calling the listener callbacks. If that is the case, I don't know if it gets started during unit tests, but either way, it would be ideal to be able to explicitly call its run method or whatever during unit tests and not have to deal with the dragons of multithreading

virtuald:
It would get started automatically when NT is initialized. There's a function you can call in the low level api (waitFor... something) that will return once all callbacks have been successfully dispatched.

PR #339, opened 2017-11-30
add more unit tests for relay

ariovistus:


wpilib/tests/test_relay.py
```diff
@@ -155,13 +170,60 @@ def test_relay_get(dir, value, fwd, rev, wpilib, relay_data):
...
+
+def test_relay_isAlive1(relay):
+    relay.setSafetyEnabled(False)
+    assert relay.isAlive()
+    relay.setSafetyEnabled(True)
+    assert not relay.isAlive()
+
+
+def test_relay_isAlive2(wpilib):
+    with patch('hal_impl.functions.hooks', new=SimHooks()) as hooks:
```
virtuald: Thanks for adding tests! However, it would be better if the sim hooks was a fixture in conftest.py instead of a standalone file:

```
@pytest.fixture
def sim_hooks(hal_data):
    with patch('hal_impl.functions.hooks', new=SimHooks()) as hooks:
        yield hooks
```

wpilib/tests/test_relay.py
```diff
@@ -155,13 +170,60 @@ def test_relay_get(dir, value, fwd, rev, wpilib, relay_data):
...
+
+        relay.setSafetyEnabled(True)
+        hooks.time = 1.0
+        assert not relay.isAlive()
+        relay.feed()
+        assert relay.isAlive()
+    
+
+def test_relay_stopMotor(relay, wpilib):
+    relay.set = MagicMock()
```
virtuald: In general, my preference would be that we avoid MagicMock-based tests and use the hal_data interfaces instead. I realize there are a bunch of them already, but a lot of those were created before the simulated hal existed and we've never gotten around to changing them. The advantage (in my opinion) of using the hal_data stuff is that it tests our hal boundaries too in addition to testing the wpilib stuff (and there is some typechecking done there as well). Where it's difficult to use, we should improve (or make helpers) so that it's easier for others to use too.

However, if you don't want to change the ones you've already done, that's fine, I will still merge these.

wpilib/tests/test_relay.py
ariovistus: with this syntax, when does hal_impl.functions.hooks get restored to its original value?

wpilib/tests/test_relay.py
ariovistus: I agree, but in this particular case, set's effects are a bit opaque, and test_relay_set covers it thoroughly

wpilib/tests/test_relay.py
virtuald: If you use `@pytest.fixture(scope='function')` it will get restored after each test has ran.

wpilib/tests/test_relay.py
virtuald: Fair enough.

wpilib/tests/test_relay.py
ariovistus: seems like my sim hooks breaks some tests in test_utility, does the fixture always get initialized, or only when it is requested?

virtuald: Fixtures only are initialized when requested.

PR #343, opened 2017-12-02
Update HAL to latest

virtuald:


hal-base/hal/functions.py
```diff
@@ -32,6 +33,9 @@
 from hal_impl.fndef import _RETFUNC, _THUNKFUNC, _VAR, _dll, sleep
 from hal_impl import __hal_simulation__
 
+# This monkeypatch allows us to treat ctypes values like bytes
+C.c_char_p.decode = lambda self, encoding: self.value.decode(encoding)
```
virtuald: I haven't actually tried this on the RoboRIO, but seems to work locally.

hal-sim/hal_impl/functions.py
```diff
@@ -1850,5 +1863,14 @@ def clearAllPCMStickyFaults(module, status):
     status.value = 0
     return False
 
+def setOneShotDuration(solenoidPortHandle, durMS, status):
+    status.value = 0
+    hal_data['pcm'][solenoidPortHandle.module][solenoidPortHandle.pin]['one_shot_duration'] = durMS
+
+def fireOneShot(solenoidPortHandle, status):
+    status.value = 0
+    # TODO: need to schedule a callback to implement this somehow
```
virtuald: I don't really have a good idea on how to make this work. We would need some hal mechanism to schedule callbacks after some specified time.

wpilib/tests/test_driverstation.py
```diff
@@ -257,3 +257,21 @@ def test_InTest(dsmock):
     dsmock.InTest(False)
     assert not dsmock.userInTest
 
+
+# HAL-only tests
```
virtuald: These only vaguely belong here, but there's not really a better place for them?

PR #345, opened 2017-12-02
Update hal digital files

Twinters007:


hal-sim/hal_impl/functions.py
```diff
@@ -856,7 +856,24 @@ def freeDIOPort(dioPortHandle):
...
+                hal_data["mxp"][i]["initialized"] = True
+                handle = i
+                break
+        if not port['initialized']:
+            dio = _initport('dio', i, status)
+            handle = i
+            break
+
+    if dio is None:
+        return 0
```
auscompgeek: I'm pretty sure this will throw an error in the HAL base.

hal-sim/hal_impl/functions.py
```diff
@@ -856,7 +856,24 @@ def freeDIOPort(dioPortHandle):
...
+        if i >= kNumDigitalHeaders:
+            if not hal_data["mxp"][i]["initialized"]:
+                dio = _initport('dio', i, status)
+                hal_data["mxp"][i]["initialized"] = True
+                handle = i
+                break
+        if not port['initialized']:
+            dio = _initport('dio', i, status)
+            handle = i
+            break
```
auscompgeek: I still don't understand how this matches the athena implementation.

Also, the fact that you have duplicated code smells like you're doing something terribly wrong here.

Twinters007:
@auscompgeek @virtuald one of you might want to tackle this since my understanding of upstream is not quite there yet

PR #346, opened 2017-12-02
Update sensor base and interruptable sensor base

twinters007:


wpilib/wpilib/interruptablesensorbase.py
```diff
@@ -70,21 +67,17 @@ def allocateInterrupts(self, watcher):
             where the user program will have to explicitly wait for the interrupt
             to occur.
         """
-        if self.interrupt is not None:
-            raise ValueError("The interrupt has already been allocated")
 
         self.isSynchronousInterrupt = watcher
-        self._interrupt = hal.initializeInterrupts(watcher)
-        self._interrupt_finalizer = weakref.finalize(self, hal.cleanInterrupts,
```
virtuald: Is there a good reason to get rid of the finalizer?

Twinters007: Only because it wasn't in upstream

virtuald: This would be where we should have comments saying 'this is python specific'... we should also document the finalizer pattern in the porting guide too I suppose.

The reason for having a finalizer is so that when the object is garbage collected, the hal resources will be cleaned up automatically.

@PeterJohnson you had added these originally, if the program exits without cleaning up hal resources, do they get cleaned up by the FPGA automatically, or does that cause issues?

auscompgeek:
Please fix broken tests.

wpilib/wpilib/interruptablesensorbase.py
```diff
@@ -71,21 +77,18 @@ def allocateInterrupts(self, watcher):
             where the user program will have to explicitly wait for the interrupt
             to occur.
         """
-        if self.interrupt is not None:
-            raise ValueError("The interrupt has already been allocated")
```
auscompgeek: Should this have been removed?

wpilib/wpilib/interruptablesensorbase.py
```diff
@@ -73,19 +79,17 @@ def allocateInterrupts(self, watcher):
...
+        self._interrupt_finalizer = weakref.finalize(self, hal.cleanInterrupts, self._interrupt)
 
     def cancelInterrupts(self):
         """Cancel interrupts on this device. This deallocates all the
         chipobject structures and disables any interrupts.
         """
         if self.interrupt is None:
             raise ValueError("The interrupt is not allocated.")
-        self._interrupt_finalizer()
+        hal.cleanInterrupts(self.interrupt)
```
auscompgeek: I wonder what the implications of this change are.

twinters007: i think i was supposed to keep that

twinters007:
I think I rebased or something...still not sure about that finalizer line though

auscompgeek:
I'm not sure about that either.

wpilib/wpilib/interruptablesensorbase.py
virtuald: Yes, you should have kept it, otherwise it'll potentially be called twice.

virtuald: ... so put it back?

auscompgeek:
Shouldn't there also be a free method in here as well?

wpilib/wpilib/interruptablesensorbase.py
twinters007: Oh. I'm seeing this comment now

wpilib/wpilib/interruptablesensorbase.py
```diff
@@ -86,6 +95,7 @@ def cancelInterrupts(self):
         if self.interrupt is None:
             raise ValueError("The interrupt is not allocated.")
         self._interrupt_finalizer()
+        hal.cleanInterrupts(self.interrupt)
```
auscompgeek: Won't this get called by the finalizer?

PR #347, opened 2017-12-05
add unit tests for pid controller

ariovistus:
also, found why the new networktables impl was throwing null refs - the tests would sometimes free networktables first and then free the wpilib stuff.

### DIFF
--- a/wpilib/tests/conftest.py
+++ b/wpilib/tests/conftest.py
@@ -45,7 +45,7 @@ def hal_data(_module_patch):
 
 
 @pytest.fixture(scope="function")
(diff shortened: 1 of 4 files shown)


auscompgeek:
Something gives me the impression that the tests are correct in telling you that `wpilib._impl.utils` doesn't have a `print` attribute :smile:

I don't think we actually care about the format of the argument list matcher messages, and in fact there is a TODO comment in there about giving the user a better error message.

We definitely care about `match_arglist` raising an exception though.  I wonder whether we have a test specifically for that.

ariovistus:
damn you, python 3.4

PR #350, opened 2017-12-06
Add sphinx build to travis-ci

virtuald:


wpilib/wpilib/buttons/networkbutton.py
```diff
@@ -15,13 +15,13 @@
...
 class NetworkButton(Button):
-    """A :class:`.Button` that uses a :class:`NetworkTable` boolean field."""
+    """A :class:`.buttons.Button` that uses a :class:`.NetworkTable` boolean field."""
 
     def __init__(self, table, field):
         """Initialize the NetworkButton.
 
-        :param table: the NetworkTable instance to use, or the name of the
-        table to use.
+        :param table: the :class:`.NetworkTable` instance to use, or the name of the
```
auscompgeek: Shouldn't this be `networktables.networktable.NetworkTable`?

virtuald: That seems like an intersphinx bug, the ```:class:`.NetworkTable` ``` should just work.

Issue #157, opened 2015-12-16
Add PEP 0484 type hinting?


virtuald:
This could be useful to combine with @computer-whisperer's API checkers. Or, pyfrc could automatically run it as part of tests before deploying code -- that would be really cool. 

See http://mypy-lang.org/ and https://www.python.org/dev/peps/pep-0484/

CrazyPython:
@virtuald 

This needs to be implemented NOW. It really helps my IDE a lot. Seriously.

virtuald:
I will accept pull requests. 👍

auscompgeek:
I feel like possibly taking this up. I'm just thinking, should we use `*.pyi` stub files, or should we just annotate the source?

Twinters007:
If you annotate the source it'll show up if you aren't using an IDE and you can view them easily just by going into the file.

virtuald:
My opinion is that it should be done in the source, otherwise it'll never get updated.

You can most likely use the API checker stuff to do a lot of the work for you.

auscompgeek:
We should also make devtools/convert_javadoc.html convert the type information to type hints too.

virtuald:
That's a good idea.

auscompgeek:
Just something I've noticed: full-blown type hints don't seem to look very nice when rendered in the docs. They might look nicer with forward-references, but I haven't had a chance to try that.

nhubbard:
@virtuald I am working on this in my fork. The effort is few and far in between, but it will eventually be pull-requested into the master branch.

virtuald:
@nhubbard awesome. Probably better to send the PRs in batches instead of the whole thing all at once.

virtuald:
I think this is mostly done?

PR #351, opened 2017-12-06
Update RobotBase for 2018

auscompgeek:
Also change this to use our NetworkTables global, which points to the
default NetworkTablesInstance.

As a bonus, also use type hints, to aid static analysis tools that don't
understand Sphinx type field directives.

### DIFF
--- a/wpilib/wpilib/robotbase.py
+++ b/wpilib/wpilib/robotbase.py
@@ -1,4 +1,4 @@
-# validated: 2016-12-31 JW 8f67f2c24cb9 edu/wpi/first/wpilibj/RobotBase.java
+# validated: 2017-12-06 DV dd7563376bf6 edu/wpi/first/wpilibj/RobotBase.java
 #----------------------------------------------------------------------------
 # Copyright (c) FIRST 2008-2012. All Rights Reserved.
 # Open Source Software - may be modified and shared by FRC teams. The code
@@ -7,8 +7,7 @@
 #----------------------------------------------------------------------------
 
 import hal
-import warnings
-from networktables import NetworkTablesInstance
+from networktables import NetworkTables
 
 import logging
 logger = logging.getLogger('robotpy')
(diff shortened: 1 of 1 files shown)


virtuald:
It would be cool if we could run some of those tools as part of the tests.

wpilib/wpilib/robotbase.py
```diff
@@ -37,98 +36,85 @@ class RobotBase:
...
         """Free the resources for a RobotBase class."""
-        # TODO: delete?
         pass
 
     @staticmethod
-    def isSimulation():
-        """
-            :returns: If the robot is running in simulation.
-            :rtype: bool
+    def isSimulation() -> bool:
```
Twinters007: are we gonna start adding type hinting to stuff?

wpilib/wpilib/robotbase.py
```diff
@@ -37,98 +36,85 @@ class RobotBase:
...
 
     @staticmethod
-    def isReal():
-        """
-            :returns: If the robot is running in the real world.
-            :rtype: bool
+    def isReal() -> bool:
+        """Get if the robot is real.
+
+        :returns: If the robot is running in the real world.
```
virtuald: ... returns True?

wpilib/wpilib/robotbase.py
```diff
@@ -160,9 +146,9 @@ def main(robot_cls):
             robot = robot_cls()
         except:
             from .driverstation import DriverStation
-            DriverStation.reportError("ERROR Could not instantiate robot", True)
-            logger.exception("Robots don't quit!")
-            logger.exception("Could not instantiate robot "+robot_cls.__name__+"!")
+            DriverStation.reportError("Unhandled exception instantiating robot " + robot_cls.__name__, True)
+            DriverStation.reportWarning("Robots should not quit, but yours did!")
```
Twinters007: Took them long enough to change this 🙏

wpilib/wpilib/robotbase.py
virtuald: There is an issue that states we should do it. Adding it gradually is fine.

wpilib/wpilib/robotbase.py
Twinters007: ?

auscompgeek: I'm just mirroring what upstream have. ¯\\\_(ツ)_/¯

wpilib/wpilib/robotbase.py
auscompgeek: Per #157, yes.

Whether we'll depend on the `typing` module whilst we still support Python 3.4 is an open question however.  But that's a discussion for over there.

Twinters007: What if we made a rash decision to not support 3.4 in 2018?

wpilib/wpilib/robotbase.py
Twinters007: Why would it only return true?

auscompgeek:
> It would be cool if we could run some of those tools as part of the tests.

Aside from mypy, all the tools I'm thinking of are pretty much IDEs and IDE tooling.

PR #352, opened 2017-12-07
implement SendableBuilderImpl

ariovistus:


wpilib/wpilib/sendablebuilder.py
```diff
@@ -0,0 +1,234 @@
...
+    def addBooleanProperty(self, key, getter, setter):
+        """
+        Add a boolean property.
+   
+        :param key:     property name
+        :type key: str
+        :param getter:  getter function (returns current value)
+        :type getter: typing.Callable[[], bool]
+        :param setter:  setter function (sets new value)
+        :type setter: typing.Callable[[networktables.entry.NetworkTableEntry, str, bool, Any], None]
```
ariovistus: need guidance on the right way to express these

auscompgeek: the right way = not in the docstring

ariovistus: I though 3.4 compat prevents it from going elsewhere

virtuald: Is there a compelling reason to support 3.4 still?

auscompgeek: there's a backport of the typing module in either case

virtuald: Honestly, I would say pick one of the two:

* Use the type annotations
* Build the documentation locally, and change the type to match what you would want it to look like as sphinx renders it. I'm not too picky about it... but it seems like this wouldn't render very well?

wpilib/wpilib/sendablebuilder.py
```diff
@@ -0,0 +1,215 @@
...
+        """
+        self.table = table
+
+    def getTable(self):
+        """
+        Get the network table.
+
+        :returns: The network table
+        :rtype: networktables.networktable.NetworkTable
+        :rtype: :class:`networktables.networktable.NetworkTable`
```
auscompgeek: …

wpilib/wpilib/sendablebuilder.py
```diff
@@ -0,0 +1,215 @@
...
+        state.  This is called when entering and exiting Live Window mode.
+
+        :param func:    function
+        """
+        self.safeState = func
+
+    def setUpdateTable(self, func):
+        """
+         Set the function that should be called to update the network table
+        for things other than properties.  Note this function is not passed
```
auscompgeek: Indentation doesn't match.

wpilib/wpilib/sendablebuilder.py
ariovistus: wow, typing generates some bizarre errors in places I didn't touch. guess we'll have to go with option 2. how do I build the documentation?

ariovistus:
setter signature is wrong

wpilib/wpilib/sendablebuilder.py
virtuald: Sorry, missed your question. Install sphinx and sphinx-rtd-theme, and do `make html`. If you're on Windows, not sure if that will actually work however.

PR #353, opened 2017-12-08
update sendable, implement SendableBase.java

ariovistus:


wpilib/wpilib/sendablebase.py
```diff
@@ -0,0 +1,96 @@
+import threading
```
auscompgeek: Where's the git-source-track line?

wpilib/wpilib/sendablebase.py
```diff
@@ -0,0 +1,96 @@
...
+from .sendable import Sendable
+from .livewindow import LiveWindow
+from ._impl.utils import match_arglist
+
+
```
auscompgeek: Define `__all__`, otherwise the `import *` will pick up everything that's been imported.

wpilib/wpilib/sendablebase.py
ariovistus: added

wpilib/wpilib/sendable.py
```diff
@@ -62,3 +62,20 @@ def initSendable(self, builder):
...
+        if hasattr(self, "updateTable"):
+            self.updateTable()
+
+    # todo: remove
+    def getSmartDashboardType(self):
+        """
+            :returns: the string representation of the named data type that
+                will be used by the smart dashboard for this sendable
+        """
+        raise NotImplementedError
```
auscompgeek: Eh. Just update everything that depends on this now rather than later.

ariovistus: the last time I did that dustin made me split it up

auscompgeek: that was because you had completely unrelated changes, not because of the PR's pure size

ariovistus: At the time that I was doing the development, it appeared to be necessary to squish them together to keep the tests happy. not the unittests, the integrationy tests that I guess come from examples?
The same tests that bomb when these functions are removed. I think there are many components that depend on sendable, perhaps the majority of the outstanding changes. Or perhaps the number of additional changes that need to be done is relatively small, but I don't know and I don't feel discovering the subset of changes that need to be made just to keep the tests happy is a valuable use of anybody's time.
This way, the tests are happy, not a lot of effort has been expended, we get this in, and everybody can work on updating their files.

virtuald: I think for things that the tests are tricky, we just need to state that they're tricky, and merge them quickly. I don't want tests not working for extended periods of time, as it makes it difficult to figure out whether the changes are 'bad' or not.

Large/small aren't particularly important descriptors of a PR -- the key metrics I care about when merging a PR is 'is it easy for me to reason about the changes' and 'do the tests pass'. The first one is much more important than the second -- particularly if the tests are in fact failing. This was what the problem with the large PR you had previously -- not only was it large, but the tests weren't passing and I didn't want to deal with it. Once the PR was broken up, it was (relatively) trivial to figure out why it was broken.

Anyways, carry on.

virtuald: Also, having the tests fail can be acceptable if (a) it's well understood why they are failing and (b) is pointed out in the PR and (c) is promised to be fixed soonish.

PR #354, opened 2017-12-08
Validate LinearDigitalFilter

auscompgeek:
Upstream changed their CircularBuffer helper class to be more idiomatic, but we just use collections.deque.

(Maybe we want to reconsider this? Their CircularBuffer is array-backed and doesn't shuffle elements around, so it'd be more efficient than a doubly-linked list, right?)


virtuald:
Yeah, but a deque is implemented in C, so it'll beat any python implementation we write?

auscompgeek:
Maybe; we should do some timing tests. It'd likely be more memory efficient though.

auscompgeek:
Let's table this discussion for later. Getting a working update to teams is priority.

PR #357, opened 2017-12-10
update livewindow.py

ariovistus:
also update ultrasonic
also update inheritance chains


wpilib/wpilib/sensorbase.py
```diff
@@ -163,6 +163,9 @@ def getDefaultSolenoidModule():
         """
         return SensorBase.defaultSolenoidModule
 
+    def __init__(self):
+        super().__init__()
+
```
auscompgeek: Why have you added this useless constructor?

ariovistus: SendableBase.__init__ must be called

virtuald: Strictly speaking, if you omit the constructor, then super will be called instead. But, that's fine.

virtuald:
Looks like your tests are failing because of docstring indentation and something to do with AnalogPotentiometer.setName() ?

ariovistus:
more things that need to extend SendableBase. When I try to import SendableBase from command/subsystem.py, it errors with 

E   ImportError: cannot import name 'SendableBase'

out of `from ..sendablebase import SendableBase`

any idea what that's about?

wpilib/wpilib/doublesolenoid.py
```diff
@@ -111,9 +111,9 @@ def __init__(self, *args, **kwargs):
                                             self.forwardHandle, self.reverseHandle)
 
     def free(self):
-        """Mark the solenoid as freed."""
-        LiveWindow.removeComponent(self)
-        
+        """
+        Mark the solenoid as freed.
+        """
```
auscompgeek: Uh, why did you reformat this docstring?

wpilib/wpilib/livewindow.py
```diff
@@ -22,45 +33,28 @@ class LiveWindow:
...
     actuators on the LiveWindow."""
 
-    sensors = set()
-    #actuators = set()
     components = {}
-    livewindowTable = None
-    statusTable = None
+    liveWindowTable = NetworkTablesInstance.getDefault().getTable("LiveWindow")
+    statusTable = liveWindowTable.getSubTable(".status")
+    enabledEntry = statusTable.getEntry("LW Enabled")
```
auscompgeek: nononononononononono

auscompgeek:
> more things that need to extend SendableBase. When I try to import SendableBase from command/subsystem.py, it errors with
> 
> E ImportError: cannot import name 'SendableBase'

Error messages aren't very useful without the stack trace.

I see you've pushed code which fails with this. The stack trace will tell you why.

wpilib/wpilib/livewindow.py
ariovistus: what are the semantics of _reset? I assumed it was to recreate the initial state. What's the appropriate thing to do here?

wpilib/wpilib/doublesolenoid.py
ariovistus: it annoyed me

wpilib/wpilib/livewindow.py
auscompgeek: this is at the class level, not in any method defined — I'm pretty sure this will cause NT to be initialised

wpilib/wpilib/doublesolenoid.py
auscompgeek: well, this is annoying me now 😜

(see PEP 257)

wpilib/wpilib/servo.py
```diff
@@ -48,7 +48,6 @@ def __init__(self, channel):
                    self.getChannel())
 
     def free(self):
-        LiveWindow.removeComponent(self)
         super().free()
```
auscompgeek: Since this is now just a super call, just remove the method?

wpilib/wpilib/relay.py
```diff
@@ -26,7 +26,7 @@ def _freeRelay(*handles):
         if handle:
             hal.freeRelayPort(handle)
 
-class Relay(SensorBase, LiveWindowSendable, MotorSafety):
+class Relay(SensorBase, MotorSafety):
```
ariovistus: arg, SendableBase

wpilib/wpilib/doublesolenoid.py
ariovistus: ugh, I suppose next you're going to tell me there's a convention against a comma following the last item in an array literal

virtuald: I actually prefer adding a comma to the end of an array literal. 👍

wpilib/wpilib/livewindow.py
virtuald: The `reset` function is probably misnamed -- it probably should be called 'free'. The primary purpose of the reset function is to ensure that wpilib is correctly reinitialized during unit tests (all of the reset functions are called during test... teardown? or startup. Can't remember).

virtuald: And yeah, we shouldn't be initializing NT here, that potentially causes problems. We'll need to find a different way to initialize the class variable (maybe via a get function that initializes it if not present? Look to SmartDashboard, I think it does a similar thing).

auscompgeek:
I would suggest using `@classmethod` instead of `@staticmethod`. Should also make sure git-source-track lines are updated if applicable, and maybe update the old LiveWindow code as well.

Thanks for putting the effort into this.

virtuald:
I don't think it makes sense to take time update git source-track here.. this is touching too many things. Let the next PRs deal with that I think.

auscompgeek:
Just another note: should probably have `warnings` for the deprecated LiveWindow methods.

virtuald:
Isn't `.. deprecated` sufficient? I believe that shows up in the docs.

auscompgeek:
I think it might be nice to have a warning at runtime. In the past teams have had to override the IterativeRobot.test\* methods and call `LiveWindow.run()` themselves.

Twinters007:
I added `warnings.warn` calls to other deprecated stuff. Docs pick up the `.. deprecated` but IDEs don't I don't think (pycharm doesn't)

wpilib/wpilib/livewindow.py
```diff
@@ -185,19 +183,123 @@ def addActuatorModuleChannel(moduleType, moduleNumber, channel, component):
...
-            del(LiveWindow.components[component])
+        with LiveWindow.mutex:
+            component = LiveWindow.components.get(sendable, None)
+            if component is not None:
+                component.telemetryEnabled = False
+
+    @staticmethod
+    def disableAllTelemetry():
+        """
+        Disable ALL telemetry
```
Twinters007: hehehe in case the method name wasn't enough, here it is in all caps just to be sure

wpilib/wpilib/livewindow.py
```diff
@@ -117,10 +103,13 @@ def addSensor(subsystem, name, component):
         :param name: The name of this component.
         :param component: A LiveWindowSendable component that represents a
             sensor.
+
+        .. deprecated:: 2018.0.0
+            Use Sendable.setName(str, str) instead.
```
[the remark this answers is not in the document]
ariovistus: yup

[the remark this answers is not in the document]
Twinters007: We deal with that by using keyword arguments with default values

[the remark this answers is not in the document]
ariovistus: I meant in the docs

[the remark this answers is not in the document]
virtuald: @ariovistus what platform are you on? building the docs should be pretty easy...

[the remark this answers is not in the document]
ariovistus: nvm, looks trivial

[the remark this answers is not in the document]
ariovistus: hope it isn't supposed to hyperlink, cuz it ain't doin that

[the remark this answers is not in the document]
virtuald: Yes, it's supposed to hyperlink. I would try 

```
:meth:`.Sendable.setName`
```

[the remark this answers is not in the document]
ariovistus: is there a way to do that and keep the parameters?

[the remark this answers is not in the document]
virtuald: I don't think so.

wpilib/wpilib/livewindow.py
```diff
@@ -22,48 +33,49 @@ class LiveWindow:
...
+    startLiveWindow = False
     liveWindowEnabled = False
-    firstTime = True
+    telemetryEnabled = True
+    mutex = threading.RLock()
+
+    @classmethod
+    def liveWindowTable(cls):
+        if cls._liveWindowTable is None:
+            cls._liveWindowTable = NetworkTablesInstance.getDefault().getTable("LiveWindow")
```
auscompgeek: Just a note: you can use `NetworkTables` instead of `NetworkTablesInstance.getDefault()`.

wpilib/wpilib/livewindow.py
```diff
@@ -75,77 +87,75 @@ def setEnabled(enabled):
...
         """Add a Sensor associated with the subsystem and with call it by the
         given name.
 
         :param subsystem: The subsystem this component is part of.
         :param name: The name of this component.
         :param component: A LiveWindowSendable component that represents a
             sensor.
+
+        .. deprecated:: 2018.0.0
+            Use :py:meth:`Sendable.setName(subsystem, name)` instead.
```
auscompgeek: So I believe by listing the arguments, this doesn't actually hyperlink to the relevant docs.

You can leave off the `:py` at the beginning, the Python domain is the default here.

wpilib/wpilib/livewindow.py
```diff
@@ -1,13 +1,24 @@
-# validated: 2015-12-24 DS 6d854af edu/wpi/first/wpilibj/livewindow/LiveWindow.java
-
-from networktables import NetworkTables
-from .command import Scheduler
+# validated: 2017-12-09 EN f9bece2ffbf7 edu/wpi/first/wpilibj/livewindow/LiveWindow.java
+import threading
```
auscompgeek: Hmm. Weird how we don't have the copyright notice here.

wpilib/wpilib/livewindow.py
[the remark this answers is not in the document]
ariovistus: SendableBase.html doesn't get generated for me. Any tricks?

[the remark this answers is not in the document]
auscompgeek: Ah. It's because SendableBase doesn't have a docstring. Give the class a docstring and it should get generated.

wpilib/wpilib/sendablebase.py
```diff
@@ -15,6 +15,9 @@
 
 
 class SendableBase(Sendable):
+    """
+    """
```
auscompgeek: Could you please put a sensible docstring here? This is intended to be documented, so uh.

Also, you might like to send a PR upstream to fix their docs. Or I guess just use whatever upstream have right now. (Python teams being just as confused as Java teams isn't really our problem, right?)

ariovistus: Oop, there was a docstring

PR #358, opened 2017-12-12
update servo for 2018 some more

ariovistus:


wpilib/tests/test_servo.py
```diff
@@ -43,6 +43,9 @@ def test_servo_angle(param, expected, expectedAngle, servo, servo_data):
     assert servo.getAngle() == expectedAngle
 
 
-def test_servo_getservoanglerange(servo):
-    assert servo.getServoAngleRange() == pytest.approx(180, 0.01)
+def test_servo_getServoAngleRange(servo):
```
Twinters007: it's in test_servo, do we need `test_servo_getServoAngleRange` instead of just `test_getServoAngleRangle`?

ariovistus: well, I'm used to unit test reports that only show method name, though that probably isn't relevant here. It seemed to be the convention.. in some places.. so I went with it

PR #359, opened 2017-12-13
update power distribution panel for 2018 some more

ariovistus:


wpilib/wpilib/powerdistributionpanel.py
```diff
@@ -98,29 +97,13 @@ def clearStickyFaults(self):
...
-            for channel in range(16):
-                self.chanEntry[channel].setDouble(self.getCurrent(channel))
-        if self.voltageEntry is not None:
-            self.voltageEntry.setDouble(self.getVoltage())
-        if self.totalCurrentEntry is not None:
-            self.totalCurrentEntry.setDouble(self.getTotalCurrent())
+    def initSendable(self, builder):
+        builder.setSmartDashboardType("PowerDistributionPanel")
+        for chan in range(self.kPDPChannels):
+            builder.addDoubleProperty("Chan%s" % (chan,), self._getCurrentOf(chan), None)
```
ariovistus: is there a more idiomatic way to do this in python?

PeterJohnson: Doesn't this immediately call _getCurrentOf?  You want to pass it as a callable, so this should likely be a lambda instead.

ariovistus: _getCurrentOf returns a lambda. python doesn't have final, so this is to bind the current value of chan to the lambda.

virtuald: yeah, that's probably as good as you're going to get. Worth adding a comment.

auscompgeek: There's always the `functools.partial` idiom for this. Otherwise I've seen `lambda x=x: ...` in the wild, but it reads weirdly to me.

auscompgeek:
Update the git-source-track header?

ariovistus:
oh hey, that terminal's still open

wpilib/wpilib/powerdistributionpanel.py
```diff
@@ -98,29 +98,9 @@ def clearStickyFaults(self):
...
-            for channel in range(16):
-                self.chanEntry[channel].setDouble(self.getCurrent(channel))
-        if self.voltageEntry is not None:
-            self.voltageEntry.setDouble(self.getVoltage())
-        if self.totalCurrentEntry is not None:
-            self.totalCurrentEntry.setDouble(self.getTotalCurrent())
+    def initSendable(self, builder):
+        builder.setSmartDashboardType("PowerDistributionPanel")
+        for chan in range(self.kPDPChannels):
+            builder.addDoubleProperty("Chan%s" % (chan,), partial(self.getCurrent, chan), None)
```
auscompgeek: BTW, you could write `'Chan%d' % chan` instead; you know `range()` is only going to give you integers.

ariovistus: force of habit

Twinters007: also why add it as a tuple with one element?

auscompgeek: @Twinters007 what do you mean?

Twinters007: `(chan,)`

auscompgeek: What do you mean by "add" though? I know that's a 1-tuple…

Twinters007: Oh I mean add it to the string via formatting

auscompgeek: Common safeguard in case the single thing you're formatting in happens to be a tuple.

Twinters007: Really? Huh. But we know it's gonna be an int here, no?

auscompgeek: Yes, thank you for repeating what I said 10 comments ago.

PR #360, opened 2017-12-13
update relay for 2018 some more

ariovistus:


wpilib/wpilib/relay.py
```diff
@@ -71,6 +72,14 @@ class Value:
             "Reverse": kReverse,
         }
 
+        @classmethod
+        def getPrettyValue(cls, num):
```
Twinters007: can we get docstrings for these two?

auscompgeek: I think these should just be changed to enums.

virtuald: Yeah... these should probably be `IntEnum`.

ariovistus: ok. thought there was a reason we couldn't use them

wpilib/wpilib/relay.py
```diff
@@ -57,21 +59,18 @@ class Value:
...
-            "Off": kOff,
-            "On": kOn,
-            "Forward": kForward,
-            "Reverse": kReverse,
-        }
-
-    class Direction:
+        @classmethod
+        def getPrettyValue(cls, num):
+            return cls(num).name[1:]
```
auscompgeek: I think you can make this a normal method instead, like upstream. I would personally prefer a property, but 🤷‍♂️.

wpilib/wpilib/relay.py
```diff
@@ -266,30 +265,16 @@ def setDirection(self, direction):
         if self.direction == direction:
             return
         
-        if direction not in [self.Direction.kBoth,
-                             self.Direction.kForward,
-                             self.Direction.kReverse]:
+        if direction not in self.Direction :
```
auscompgeek: Firstly, there is an extra space before the `:`.

Secondly, this breaks users who are currently passing integer constants rather than referencing `Relay.Direction.foo`. I don't think we want to break that until after a year of explicitly deprecating that usage.

wpilib/wpilib/relay.py
```diff
@@ -57,21 +59,18 @@ class Value:
...
-        }
-
-    class Direction:
+        @classmethod
+        def getPrettyValue(cls, num):
+            return cls(num).name[1:]
+
+        @classmethod
+        def getValueOf(cls, name):
+            match = [x for x in cls if x.name == 'k' + name]
```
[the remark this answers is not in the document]
ariovistus: default value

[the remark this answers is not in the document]
auscompgeek: Yes. Catch the KeyError it raises.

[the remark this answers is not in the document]
auscompgeek: Or you could use `getattr()` if you really wanted. Or `cls._member_map_.get()` (although probably not a good idea).

wpilib/wpilib/relay.py
ariovistus: aha, I assumed get was returning ints

wpilib/wpilib/relay.py
```diff
@@ -266,30 +261,16 @@ def setDirection(self, direction):
         if self.direction == direction:
             return
         
-        if direction not in [self.Direction.kBoth,
-                             self.Direction.kForward,
-                             self.Direction.kReverse]:
+        if direction not in list(self.Direction):
```
auscompgeek: Heh. I guess that works.

PR #362, opened 2017-12-15
update pidcontroller for 2018 some more

ariovistus:
also, add a few unit tests

### DIFF
--- a/wpilib/wpilib/pidcontroller.py
+++ b/wpilib/wpilib/pidcontroller.py
@@ -1,4 +1,4 @@
-# validated: 2017-11-19 EN 14fcf3f2f0d9 edu/wpi/first/wpilibj/PIDController.java
+# validated: 2017-12-13 EN f9bece2ffbf7 edu/wpi/first/wpilibj/PIDController.java
 #----------------------------------------------------------------------------
 # Copyright (c) FIRST 2008-2016. All Rights Reserved.
 # Open Source Software - may be modified and shared by FRC teams. The code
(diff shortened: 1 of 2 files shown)


wpilib/wpilib/pidcontroller.py
```diff
@@ -141,67 +136,76 @@ def __init__(self, Kp, Ki, Kd, *args, **kwargs):
...
-                self.result = self.clamp(self.result, self.minimumOutput, self.maximumOutput)
+            result = self.clamp(result, minimumOutput, maximumOutput)
 
-                pidOutput = self.pidOutput
-                result = self.result
-                
-                self.buf.append(self.error)
+            with self.pidWriteMutex:
+                with self.mutex:
+                    enabled = self.enabled
```
ariovistus: Need eyes on this bit. It looks to me like this is equivalent to what's in upstream, but not sure.
It also looks like upstream could have just called isEnabled

ariovistus:
question for upstream: reset uses reentrant locks but is still adorned with synchronized. Is it supposed to be this way and if so why?

wpilib/wpilib/pidcontroller.py
virtuald: Yeah, this interpretation looks correct to me.

virtuald:
I don't see a synchronized marker anywhere in PIDController?

ariovistus:
hm. must have been looking at de63e1c8a1131b0d23bdc78c3a09896e4037bbbe as presented by source track.

auscompgeek:
I'm not seeing any synchronized markers either. We have 899892c1193fe21afdfb8a6f20b007e4ce609120 currently tracked by git-source-track (although we should bump that).

PR #363, opened 2017-12-16
docs: Use sphinx-autodoc-typehints

auscompgeek:
We're trying to move towards using type hints to aid tools like mypy.
This makes the type hints not look rubbish in the docs.


auscompgeek:
To clarify: anything other than a builtin type will always show the fully qualified name in the signature. Not pretty.

PR #365, opened 2017-12-16
update counter.py for 2018 some more

ariovistus:


wpilib/wpilib/counter.py
```diff
@@ -8,6 +8,7 @@
 
 import hal
 import weakref
+from enum import IntEnum
```
virtuald: In this instance, I'd probably prefer to just import enum.

wpilib/wpilib/counter.py
```diff
@@ -44,7 +45,7 @@ class Counter(SensorBase):
     .. not_implemented: initCounter
     """
 
-    class Mode:
+    class Mode(IntEnum):
```
virtuald: Hm, does ctypes convert this to an integer when it goes down into HAL? Will need to remember to test that at some point.

PR #367, opened 2017-12-16
update command for 2018 some more

ariovistus:


wpilib/wpilib/command/command.py
```diff
@@ -49,18 +49,21 @@ def __init__(self, name=None, timeout=None):
...
+            name = self.__class__.__name__
+
+        self.setName(name)
 
         # The time (in seconds) before this command "times out" (or None if no
         # timeout)
         if timeout is not None and timeout < 0:
             raise ValueError("Timeout must not be negative")
+        if timeout is None:
+            timeout = -1.0
```
auscompgeek: Hmm. Is there a reason for adding this?

wpilib/wpilib/command/subsystem.py
```diff
@@ -138,36 +133,36 @@ def getCurrentCommand(self):
...
-        
-        :returns: the name of this subsystem
+        :returns: the current command name
+        """
+        currentCommand = self.getCurrentCommand()
+        if currentCommand is not None:
+            return currentCommand.getName()
+        return ""
+
+    def addChild(self, child, name=None):
```
auscompgeek: This doesn't match the order of the optional `name` upstream. This is a new API, so we could deviate from upstream I suppose. Do we want to here, @virtuald?

wpilib/wpilib/command/subsystem.py
```diff
@@ -138,36 +133,36 @@ def getCurrentCommand(self):
...
-                self.hasDefaultEntry.setBoolean(False)
-
-            if self.currentCommand is not None:
-                self.hasCommandEntry.setBoolean(True)
-                self.commandEntry.setString(self.currentCommand.getName())
-            else:
-                self.hasCommandEntry.setBoolean(False)
+        Associate a :class:`.Sendable` with this Subsystem.
+        Update the child's name if provided
+        :param child: sendable
```
auscompgeek: Add a blank line before this.

Twinters007: And capitalize Sendable

auscompgeek: @Twinters007 File a bug upstream if you care.

Twinters007: We can differ on capitalization, no?

auscompgeek: I don't see any reason to.

Twinters007: Looks better in the docs imo.

auscompgeek: Inconsistency is bad, and _doesn't_ look good.

virtuald: I think change it to be Right here _and_ send a PR upstream.

wpilib/wpilib/command/subsystem.py
virtuald: Optional arguments have to go last, so deviation is fine.

wpilib/wpilib/command/subsystem.py
auscompgeek: We should also change the rest of them to be consistent then.

wpilib/wpilib/command/command.py
ariovistus: yes, in the java wpilib the default value is -1.0, and other methods in this class expect it to be -1.0 and break if it is None. There are tests that exercise this.

ariovistus:
blank line added, could other changes be done in a separate PR please? 

\<bikeshedding>actually, if you were to capitalize it, you're just reporting the type, so shouldn't it be

```
The :class:`.Sendable`
```
\</bikeshedding>

Twinters007:
I wanted you to capitalize it because it's like a sentence and should start with caps (imo)

ariovistus:
making that a consistently followed rule sounds like an even bigger task

wpilib/wpilib/command/command.py
Twinters007: 2 lines above you say timeout cannot be negative, yet here you set it to negative...Why not set it to -1 if it's negative?

Twinters007:
I know I've been doing it... ¯\\\_(ツ)\_/¯

ariovistus:
are these conventions being documented anywhere? seems like an explicit checklist would be handy for new contributors.

Twinters007:
Eh not really. Notice how we also fight over whitespace a lot

virtuald:
You could just document it by changing the robotpy-docs and see if the PR gets merged.

PR #368, opened 2017-12-16
update scheduler for 2018 some more

ariovistus:


wpilib/wpilib/command/scheduler.py
```diff
@@ -217,52 +217,31 @@ def enable(self):
...
-
-    def getSmartDashboardType(self):
-        return "Scheduler"
+    def _updateTable(self):
+        if self.namesEntry is not None and self.idsEntry is not None and self.cancelEntry is not None:
+            toCancel = self.cancelEntry.getDoubleArray([])
+            if toCancel:
+                for command in list(self.commandTable):
+                    for d in toCancel:
+                        if float(command.__hash__()) == d:
```
auscompgeek: Don't call magic methods like this. `hash(command)`

Probably more reliable to use `id()` though, like we did before?

Also, the conversion to float is unnecessary (`1 == 1.0`).

wpilib/wpilib/command/scheduler.py
```diff
@@ -217,52 +217,31 @@ def enable(self):
...
+                self.cancelEntry.setDoubleArray([])
+
+            if self.runningCommandsChanged:
+                number = len(self.commandTable)
+                commands = []
+                ids = []
+
+                for command in list(self.commandTable):
+                    commands.append(command.getName())
+                    ids.append(command.__hash__())
```
auscompgeek: Ditto.

wpilib/wpilib/command/scheduler.py
```diff
@@ -217,52 +217,31 @@ def enable(self):
...
-            self.namesEntry.setStringArray(self.commands)
-            self.idsEntry.setDoubleArray(self.ids)
-
-    def getSmartDashboardType(self):
-        return "Scheduler"
+    def _updateTable(self):
+        if self.namesEntry is not None and self.idsEntry is not None and self.cancelEntry is not None:
+            toCancel = self.cancelEntry.getDoubleArray([])
+            if toCancel:
+                for command in list(self.commandTable):
```
auscompgeek: Hm. I'm guessing the `list()` is required because `self.commandTable` is indirectly mutated by the loop? Should have a comment here to that effect.

If it's not, then leave off the `list()`.

wpilib/wpilib/command/scheduler.py
```diff
@@ -217,52 +217,31 @@ def enable(self):
...
-            # Set the the running commands
-            for command in self.commandTable:
-                self.commands.append(command.getName())
-                self.ids.append(id(command))
-            self.namesEntry.setStringArray(self.commands)
-            self.idsEntry.setDoubleArray(self.ids)
-
-    def getSmartDashboardType(self):
-        return "Scheduler"
+    def _updateTable(self):
```
auscompgeek: Slowly coming to the realisation that this entire method should probably be identical to the entire old updateTable method, but with less `self`.

wpilib/wpilib/command/scheduler.py
ariovistus: id sounds better but how do you mock it? float conversion is necessary, there's a unit test

wpilib/wpilib/command/scheduler.py
```diff
@@ -217,52 +217,31 @@ def enable(self):
...
             return
 
-        # Get the commands to cancel
-        self.toCancel = self.cancelEntry.getDoubleArray([])
-        if self.toCancel:
+        toCancel = self.cancelEntry.getDoubleArray([])
+        if toCancel:
             for command in self.commandTable:
-                if id(command) in self.toCancel:
+                if float(id(command)) in toCancel:
```
auscompgeek:
```python
>>> 1 in [1.0]
True
```

wpilib/wpilib/command/scheduler.py
```diff
@@ -217,52 +217,31 @@ def enable(self):
...
-            self.idsEntry = None
-            self.cancelEntry = None
+    def initSendable(self, builder):
+        builder.setSmartDashboardType("Scheduler")
+        self.namesEntry = builder.getEntry("Names")
+        self.idsEntry = builder.getEntry("Ids")
+        self.cancelEntry = builder.getEntry("Cancel")
+        builder.setUpdateTable(self.updateTable)
 
     def updateTable(self):
```
auscompgeek: Mark this as private ­(leading underscore), since this isn't intended to be exposed to users.

wpilib/wpilib/command/scheduler.py
ariovistus: -9223363293403863878 in [float(-9223363293403863878)]

auscompgeek:
```python
>>> -9223363293403863878 .bit_length()
-63
```

So that's not surprising. I don't think the heap can go that far.

ariovistus: I assume the cause is at that value the distance between consecutive numbers on the ieee754 number line is greater than 1. Don't know (or care) what it is, or where the cutoff point is, but the float conversion prevents it from being an issue

auscompgeek: The fact that this will never happen in reality also prevents it from being an issue.

PR #369, opened 2017-12-16
update sendablechooser and smartdashboard some more

ariovistus:


wpilib/wpilib/sendablechooser.py
```diff
@@ -88,19 +82,13 @@ def getSelected(self):
 
         :returns: the object associated with the selected option
         """
-        selected = None
+        selected = self.defaultChoice
         if self.tableSelected is not None:
             selected = self.tableSelected.getString(None)
         return self.map.get(selected, self.map[self.defaultChoice])
```
auscompgeek: I don't think these last two lines match upstream.

wpilib/wpilib/sendablechooser.py
```diff
@@ -88,19 +82,13 @@ def getSelected(self):
...
-            self.tableDefault = table.getEntry(self.DEFAULT)
-            self.tableSelected = table.getEntry(self.SELECTED)
-            self.tableOptions = table.getEntry(self.OPTIONS)
-            self.tableOptions.setStringArray(self.map.keys())
-            if self.defaultChoice is not None:
-                self.tableDefault.setString(self.defaultChoice)
+    def initSendable(self, builder):
+        builder.setSmartDashboardType("String Chooser")
+        builder.addStringProperty(self.DEFAULT, lambda: self.defaultChoice, None)
+        builder.addStringArrayProperty(self.OPTIONS, lambda: self.map.keys(), None)
```
auscompgeek: `self.map.keys` instead of `lambda: self.map.keys()`? `self.map` should always be the same object.

ariovistus:
I have a feeling this PR and #367 need to be combined to prevent tests from failing

virtuald:
Individually both PRs look good to me. You should be able to merge them and push to a branch on the robotpy-wpilib repo, which will trigger the travis-ci tests. If the tests pass, then we can merge.

auscompgeek:
Should also make sure that the docstrings are up to date as well.

Octopus merges are my speciality.

wpilib/wpilib/sendablechooser.py
auscompgeek: This comment was never addressed, BTW. Doesn't really matter though.

auscompgeek:
See [build #741](https://travis-ci.org/robotpy/robotpy-wpilib/builds/319461646) — tests complain about AnalogPotentiometer when the two PRs are merged.

wpilib/wpilib/sendablechooser.py
ariovistus: i think it's more robust the way it is

wpilib/wpilib/sendablechooser.py
```diff
@@ -1,23 +1,23 @@
...
 
-class SendableChooser(Sendable):
+class SendableChooser(SendableBase):
     """A useful tool for presenting a selection of options to be displayed on
-    the SmartDashboard
+    the :class:`.SmartDashboard`
 
     For instance, you may wish to be able to select between multiple
-    autonomous modes. You can do this by putting every possible Command
+    autonomous modes. You can do this by putting every possible :class:`wpilib.command.Command`
```
ariovistus: link doesn't work on this one. I tried `..command.Command` also

auscompgeek: Does `.Command` not work?

wpilib/wpilib/smartdashboard.py
```diff
@@ -53,41 +64,51 @@ def getTable(cls):
...
     def putData(cls, *args, **kwargs):
-        """Maps the specified key to the specified value in this table.
+        """
+        Two argument formats are supported: 
+        
+        key, data
+        value
+
+        Maps the specified key (name of the :class:`.Sendable` if not provided) 
+        to the specified value in this table.
```
auscompgeek: Huh. I only just noticed this. The summary line should come first — if I come here wondering what this method does, the first thing I'd want to see is what it does, not the fact that there are two ways to specify arguments.

Keep the details about the arguments next to the documentation about the arguments.

wpilib/wpilib/smartdashboard.py
```diff
@@ -64,19 +64,21 @@ def getTable(cls):
...
         The value can be retrieved by calling the get method with a key that
         is equal to the original key.
 
-        Two argument formats are supported: key, data:
+        Two argument formats are supported: 
+        
+        key, data
+        value
```
auscompgeek: hmm…

image

PR #370, opened 2017-12-16
Add checks to ensure simulation HAL is not installed on RoboRIO

### LINKED ISSUE #303: Prevent installation of simulated hal on roborio
This can lead to really difficult to debug situations. Related to https://github.com/robotpy/robotpy-ctre/issues/16

Additionally, we should output a log message or something when the robot code starts, to make it obvious that it's in simulation mode.

stmobo:
This should fix #303.

This adds some fairly simple checks to make sure that the `robotpy-hal-sim` package isn't accidentally installed onto an actual RoboRIO. This also adds logging to `wpilib.run` that indicates when a simulated HAL is in use, and warns whenever a simulated HAL is being used on a RoboRIO.

### DIFF
--- a/hal-roborio/setup.py
+++ b/hal-roborio/setup.py
@@ -7,6 +7,11 @@ import sys
 
 from setuptools import setup
 
+# Check to see if we're on a RoboRIO before potentially downloading anything
+if __name__ == '__main__' and len(sys.argv) > 1 and sys.argv[1] == 'install':
+    if not exists('/etc/natinst/share/scs_imagemetadata.ini'):
+        raise RuntimeError("This HAL should only be installed onto a RoboRIO. Perhaps try the `robotpy-hal-sim` package?")
+
 setup_dir = dirname(__file__)
 git_dir = join(setup_dir, '..', '.git')
 base_package = 'hal_impl'
--- a/hal-sim/setup.py
+++ b/hal-sim/setup.py
@@ -55,6 +55,8 @@ if __name__ == '__main__':
         if 'robotpy-hal-roborio' in installed_packages:
             raise RuntimeError("The simulation HAL cannot be installed alongside the RoboRIO HAL.")
 
+        # Another check to see if we're on a RoboRIO.
+        # NOTE: may have false positives, but it should work well enough
         if exists('/etc/natinst/share/scs_imagemetadata.ini'):
             raise RuntimeError("The simulation HAL should not be installed onto the RoboRIO. Perhaps try the `robotpy-hal-roborio` package?")
 
--- a/hal-sim/setup.py
+++ b/hal-sim/setup.py
@@ -3,6 +3,7 @@
 from os.path import dirname, exists, join
 import sys, subprocess
 from setuptools import setup
+import pip
 
 setup_dir = dirname(__file__)
 git_dir = join(setup_dir, '..', '.git')
@@ -46,10 +47,9 @@ if __name__ == '__main__':
 
         # Check to see if the RoboRIO HAL is installed before installing the
         # simulated HAL:
-        freeze_out = subprocess.check_output(['pip3', 'freeze'])
         installed_packages = [
-            line.decode().split('==')[0]
-            for line in freeze_out.splitlines()
+            dist.project_name
+            for dist in pip.utils.get_installed_distributions()
         ]
 
         if 'robotpy-hal-roborio' in installed_packages:
--- a/hal-sim/setup.py
+++ b/hal-sim/setup.py
@@ -39,26 +39,45 @@ else:
 with open(join(setup_dir, 'README.rst'), 'r') as readme_file:
     long_description = readme_file.read()
 
-setup(
-    name='robotpy-hal-sim',
-    version=__version__,
-    description='WPILib HAL layer for simulations',
-    long_description=long_description,
-    author='Peter Johnson, Dustin Spicuzza',
-    author_email='robotpy@googlegroups.com',
-    url='https://github.com/robotpy/robotpy-wpilib',
-    keywords='frc first robotics hal can',
-    packages=['hal_impl'],
-    install_requires='robotpy-hal-base==' + __version__, # is this a bad idea?
-    license="BSD License",
-    classifiers=[
-        "Development Status :: 5 - Production/Stable",
-        "Intended Audience :: Developers",
-        "Intended Audience :: Education",
-        "License :: OSI Approved :: BSD License",
-        "Operating System :: OS Independent",
-        "Programming Language :: Python :: 3.4",
-        "Programming Language :: Python :: 3.5",
-        "Topic :: Scientific/Engineering"
-    ]
-    )
+if __name__ == '__main__':
+    if len(sys.argv) > 1 and sys.argv[1] == 'install':
+        # setuptools doesn't seem to have a nice way of running checks before
+        # an install script runs, so I'm just going to do this and hope it works
+
+        # Check to see if the RoboRIO HAL is installed before installing the
+        # simulated HAL:
+        freeze_out = subprocess.check_output(['pip3', 'freeze'])
+        installed_packages = [
+            line.decode().split('==')[0]
+            for line in freeze_out.splitlines()
+        ]
+
+        if 'robotpy-hal-roborio' in installed_packages:
+            raise RuntimeError("The simulation HAL cannot be installed alongside the RoboRIO HAL.")
+
+        if exists('/etc/natinst/share/scs_imagemetadata.ini'):
+            raise RuntimeError("The simulation HAL should not be installed onto the RoboRIO. Perhaps try the `robotpy-hal-roborio` package?")
+
+    setup(
+        name='robotpy-hal-sim',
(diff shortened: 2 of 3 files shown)


wpilib/wpilib/_impl/main.py
```diff
@@ -19,23 +21,30 @@ def _log_versions():
...
         logger.info("HAL library version %s", hal_impl.version.__hal_version__)
-    
+
     # should we just die here?
     if hal.__version__ != wpilib.__version__ and \
        hal.__version__ != hal_impl.__version__:
         logger.warning("Core component versions are not identical! This is not a supported configuration, and you may run into errors!")
-        
+
+    if hal_impl.__hal_simulation__:
```
auscompgeek: I think `hal.isSimulation()` is the preferred way of spelling this.

stmobo: Ah, didn't notice that. Thanks.

stmobo:
Would it be better to check for the existence of directories or files under `/home/lvuser` instead, like `/home/lvuser/frc`?

auscompgeek:
Well, we use the same check in other places too. Keeping it is fine, as long as there's a comment noting it's not 100% reliable.

auscompgeek:
Looks good. I'd like to have others sanity check the setup.py before we merge though — in particular, we should make sure it works on Windows; I suspect the subprocess call might fail in a bunch of environments.

stmobo:
The subprocess call should work as long as `pip3` is on the PATH, and I'm pretty sure that holds for most Windows machines with Python installed by default (including mine) and on the RoboRIO (at least when using a VM).

auscompgeek:
So I believe this would call the _wrong_ pip if I were to invoke a virtualenv Python without putting it on my PATH, so that wouldn't be good.

Also, I believe `pip`/`pip3` is _not_ on PATH by default on Windows installs of Python.

stmobo:
Okay, so I did some digging and found a function in `pip.utils` that should work better for the check. This should work as long as pip is importable (though I'm not sure if that fixes anything with regards to Windows defaults).

hal-sim/setup.py
```diff
@@ -39,26 +40,46 @@
...
+            dist.project_name
+            for dist in pip.utils.get_installed_distributions()
+        ]
+
+        if 'robotpy-hal-roborio' in installed_packages:
+            raise RuntimeError("The simulation HAL cannot be installed alongside the RoboRIO HAL.")
+
+        # Another check to see if we're on a RoboRIO.
+        # NOTE: may have false positives, but it should work well enough
+        if exists('/etc/natinst/share/scs_imagemetadata.ini'):
```
virtuald: I think this is probably the only check we need. I wouldn't even bother checking for install -- what operations would a user use a roborio for with simulated Hal? Unlikely anything I suspect.

stmobo:
Okay, so I've removed the redundant package check. I just realized something, though: since the method for installing wheels is effectively just unzipping files and dumping them somewhere, doesn't that mean that `setup.py` gets skipped over completely?

auscompgeek:
We don't publish wheels.

virtuald:
I plan to merge this once I can test it. Which should be... this week.

Twinters007:
Oh looks like I didn't submit my review...oh well

auscompgeek:
I'm heading into uni right now, so I could probably try testing this today.

Twinters007:
You have a rio?

hal-sim/setup.py
```diff
@@ -39,26 +39,45 @@
...
+        install_requires='robotpy-hal-base==' + __version__, # is this a bad idea?
+        license="BSD License",
+        classifiers=[
+            "Development Status :: 5 - Production/Stable",
+            "Intended Audience :: Developers",
+            "Intended Audience :: Education",
+            "License :: OSI Approved :: BSD License",
+            "Operating System :: OS Independent",
+            "Programming Language :: Python :: 3.4",
+            "Programming Language :: Python :: 3.5",
```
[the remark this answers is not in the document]
auscompgeek: Well, that should happen to all our packages. It's not really important though…

hal-sim/setup.py
```diff
@@ -39,26 +39,35 @@
...
-        "Topic :: Scientific/Engineering"
-    ]
-    )
+if __name__ == '__main__':
+    if len(sys.argv) > 1 and sys.argv[1] == 'install':
+        # setuptools doesn't seem to have a nice way of running checks before
+        # an install script runs, so I'm just going to do this and hope it works
+
+        # NOTE: may have false positives, but it should work well enough
+        if exists('/etc/natinst/share/scs_imagemetadata.ini'):
```
Twinters007: Can you add the 'same' check to hal-roborio/setup? `if not exists(...`

auscompgeek: I can't say I can imagine anyone accidentally installing robotpy-hal-roborio on their computer, really.

hal-sim/setup.py
```diff
@@ -39,26 +39,35 @@
...
+            raise RuntimeError("The simulation HAL should not be installed onto the RoboRIO. Perhaps try the `robotpy-hal-roborio` package?")
```
auscompgeek: I believe the canonical spelling is 'roboRIO'.

hal-sim/setup.py
Twinters007: it...could happen. really only for devs who are installing from source

virtuald:
I'm not going to be able to test this with a RIO anytime soon, but I'm ok with merging this if others are.

twinters007:
I can't figure out how to install from source cause it tries to compile, but if someone can explain how to do that I can test tonight

twinters007:
```bash
admin@roboRIO-1418-FRC:~/hal-sim# ls
LICENSE.txt               MANIFEST.in               README.rst                build                     dist                      hal_impl                  robotpy_hal_sim.egg-info  setup.py
admin@roboRIO-1418-FRC:~/hal-sim# pip3 install .
Processing /home/admin/hal-sim
Requirement already satisfied: robotpy-hal-base==2017.1.2.post0.dev221 in /usr/local/lib/python3.6/site-packages (from robotpy-hal-sim==2017.1.2.post0.dev221)
Installing collected packages: robotpy-hal-sim
  Found existing installation: robotpy-hal-sim 2017.1.2
    Uninstalling robotpy-hal-sim-2017.1.2:
      Successfully uninstalled robotpy-hal-sim-2017.1.2
  Running setup.py install for robotpy-hal-sim ... error
    Complete output from command /usr/local/bin/python3.6 -u -c "import setuptools, tokenize;__file__='/var/volatile/tmp/pip-v55tmlb6-build/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-amy_12_l-record/install-record.txt --single-version-externally-managed --compile:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/var/volatile/tmp/pip-v55tmlb6-build/setup.py", line 49, in <module>
        raise RuntimeError("The simulation HAL should not be installed onto the RoboRIO. Perhaps try the `robotpy-hal-roborio` package?")
    RuntimeError: The simulation HAL should not be installed onto the RoboRIO. Perhaps try the `robotpy-hal-roborio` package?

    ----------------------------------------
  Rolling back uninstall of robotpy-hal-sim
Command "/usr/local/bin/python3.6 -u -c "import setuptools, tokenize;__file__='/var/volatile/tmp/pip-v55tmlb6-build/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-amy_12_l-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /var/volatile/tmp/pip-v55tmlb6-build/
admin@roboRIO-1418-FRC:~/hal-sim#
```
seems to work

hal-roborio/setup.py
```diff
@@ -7,6 +7,11 @@
 
 from setuptools import setup
 
+# Check to see if we're on a RoboRIO before potentially downloading anything
+if __name__ == '__main__' and len(sys.argv) > 1 and sys.argv[1] == 'install':
```
twinters007: In the hal-sim file this part is at the bottom. It needs to go there for it to work

virtuald: Why does it need to go on the bottom?

twinters007: idk. it doesn't work at the top. It's like name isn't `__main__` yet.

auscompgeek: `__name__` doesn't just arbitrarily change midway through a script. You're doing something wrong.

twinters007: `pip install .`

twinters007: Ah. `sys.argv[1]` is not 'install'. It's 'egg_info'

twinters007: virtualenv pip did some funky stuff, and I had to remove from cache

PR #373, opened 2017-12-22
port NidecBrushless.java

ariovistus:


wpilib/wpilib/nidecbrushless.py
```diff
@@ -0,0 +1,140 @@
...
+        self.isInverted = False
+        self.disabled = False
+
+        hal.report(hal.UsageReporting.kResourceType_NidecBrushless, pwmChannel)
+        self.setName("Nidec Brushless", pwmChannel)
+
+    def free(self):
+        """Free the resources used by this object."""
+        super().free()
+        self.dio.free()
```
ariovistus: right, finalizer pattern

virtuald: If you're not directly allocating a hal resource, then I don't think a finalizer is necessary. Let garbage collection do it's thing.

wpilib/wpilib/nidecbrushless.py
```diff
@@ -0,0 +1,140 @@
...
+        self.pwm.free()
+
+    def set(self, speed):
+        """ 
+        Set the PWM value.
+    
+        The PWM value is set using a range of -1.0 to 1.0, appropriately scaling the value for the FPGA.
+
+       :param speed: The speed value between -1.0 and 1.0 to set.
```
auscompgeek: Indentation doesn't match.

wpilib/wpilib/nidecbrushless.py
```diff
@@ -0,0 +1,140 @@
...
+        self.pwm.setDisabled()
+
+    def enable(self):
+        """
+        Re-enable the motor after :meth:`disable` has been called.  
+        The :meth:`set` function must be called to set a new motor speed.
+        """
+        self.disabled = False
+        
+
```
auscompgeek: One too many newlines here I think.

wpilib/wpilib/nidecbrushless.py
virtuald: @ariovistus are you converting docstrings by hand? That sounds painful, checkout `devtools/convert_javadoc.html`

PR #375, opened 2017-12-22
Update gittrack

Twinters007:


auscompgeek:
btw you forgot to bump the HAL version to go with this. patiently waiting for an update that needs the new HAL functions…

virtuald:
The HAL stuff brings in a bunch of SPI changes, so I've been putting it off.

PR #378, opened 2017-12-23
update DigitalOutput for 2018 some more

ariovistus:


wpilib/wpilib/digitaloutput.py
```diff
@@ -36,18 +37,18 @@ def __init__(self, channel):
         :param channel: the DIO channel for the digital output. 0-9 are on-board, 10-25 are on the MXP
         """
 
+        super().__init__()
         self._pwmGenerator = None
         self._pwmGenerator_finalizer = None
 
-        self.checkDigitalChannel(channel)
+        SensorBase.checkDigitalChannel(channel)
```
Twinters007: I tried making this change and someone said something before...

ariovistus: I think these changes make DigitalOutput no longer inherit from SensorBase, so it has to be this way now.

PR #380, opened 2017-12-24
Remove 3.4 from wpilib

Twinters007:


virtuald:
Fixes #380 (but need other fixes, docs, etc...)

PR #382, opened 2017-12-24
Add RobotController

virtuald:


wpilib/wpilib/robotcontroller.py
```diff
@@ -0,0 +1,197 @@
...
+# the project.
+#----------------------------------------------------------------------------
+
+import hal
+
+__all__ = ['RobotController']
+
+
+class RobotController:
+
```
auscompgeek: Does this need to be documented?

wpilib/wpilib/robotcontroller.py
```diff
@@ -0,0 +1,197 @@
...
+import hal
+
+__all__ = ['RobotController']
+
+
+class RobotController:
+    
+    @staticmethod
+    def getFPGAVersion() -> int:
```
Twinters007: yay for not supporting 3.4

auscompgeek: Function annotations were introduced in 3.0.

PR #383, opened 2017-12-24
Fix version display on Read the Docs

auscompgeek:
We weren't installing the wpilib package on RTD.  This prevented the
generation of the version file, which in turn led to the docs not
showing the version as 'master' instead of the actual version.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,3 +2,4 @@ sphinx
 sphinx-rtd-theme
 sphinx-autodoc-typehints
 git+git://github.com/robotpy/pynetworktables.git
+-e ../wpilib


auscompgeek:
inb4 this doesn't actually work on the RTD env

auscompgeek:
Should this go in some configuration on RTD itself rather than here? I can't say I completely understand how the RTD build process works.

virtuald:
I don't understand your description of the problem. What is this addressing?

auscompgeek:
conf.py grabs the version, right? This gets displayed in the RTD theme in the `<title>` and in intersphinx links at the very least. Except for this project it's always 'master' since the version.py is never generated - unlike robotpy-wpilib-utilities or pynetworktables, for example.

auscompgeek:
Reading the build logs on RTD, this might be the wrong place to fix this problem. I don't think the pwd when installing this requirements.txt file matches between Travis and RTD.

Twinters007:
Can you add a screenshot showing what it looks like now and what it'll look like with your fix?

virtuald:
It looks like the RTD configurations for pyfrc/pynetworktables does an install in the virtualenv, and this one doesn't. So.. this would probably fix it.

virtuald:
That broke the docs build:

> ../wpilib should either be a path to a local project or a VCS url beginning with svn+, git+, hg+, or bzr+

auscompgeek:
Yeah. Looks like my suspicions about the pwd confusion were right.

PR #384, opened 2017-12-24
Add Notifier

virtuald:
- Note that pyfrc will require a special hook for this to work

### DIFF
--- a/wpilib/wpilib/notifier.py
+++ b/wpilib/wpilib/notifier.py
@@ -1,0 +1,124 @@
+# validated: 2017-12-24 DS 8b7aa61091df edu/wpi/first/wpilibj/Notifier.java
+#----------------------------------------------------------------------------
+# Copyright (c) FIRST 2016-2017. All Rights Reserved.
+# Open Source Software - may be modified and shared by FRC teams. The code
+# must be accompanied by the FIRST BSD license file in the root directory of
+# the project.
+#----------------------------------------------------------------------------
+
+import threading
+
+import hal
+
+from .resource import Resource
+from .robotcontroller import RobotController
+
+class Notifier:
+    
+    def __init__(self, run: callable):
+        
+        #: The lock for the process information.
+        self._processLock = threading.RLock()
+        
+        # Notifier handle
+        self._notifier = hal.initializeNotifier()
+        
+        #: The time, in microseconds, at which the corresponding handler should be
+        #: called. Has the same zero as Timer.getFPGATime().
+        self._expirationTime = 0
+        
+        #: The handler passed in by the user which should be called at the
+        #: appropriate interval.
+        self._handler = run
+        
+        # Whether we are calling the handler just once or periodically.
+        self._periodic = False
+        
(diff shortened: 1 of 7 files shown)


wpilib/wpilib/notifier.py
```diff
@@ -0,0 +1,124 @@
...
+        
+        #: The thread waiting on the HAL alarm
+        self._thread = threading.Thread(target=self._thread, daemon=True)
+        self._thread.start()
+        
+        # python-specific
+        Resource._add_global_resource(self)
+    
+    # python-specific: this is called finalize in Java
+    def free(self):
```
Twinters007: no `-> None`?

virtuald:
One thing.. so I'm ignoring _"Private/protected functions (but NOT variables) should start with an underscore"_ ... I feel like this rule should change, and that private variables should start with an underscore. Thoughts?

Twinters007:
Yeah it never made sense to me. If they make it private their, might as well make it private here. It's also a good check to see if we're doing something wrong, or something python specific

wpilib/wpilib/notifier.py
```diff
@@ -0,0 +1,124 @@
+# validated: 2017-12-24 DS 8b7aa61091df edu/wpi/first/wpilibj/Notifier.java
+#----------------------------------------------------------------------------
+# Copyright (c) FIRST 2016-2017. All Rights Reserved.
+# Open Source Software - may be modified and shared by FRC teams. The code
+# must be accompanied by the FIRST BSD license file in the root directory of
+# the project.
+#----------------------------------------------------------------------------
+
+import threading
```
Twinters007: Since the rio has 2 cores, would it be beneficial to try and use multiprocessing instead of threading? I don't know much about it, so idk how hard it'd be to implement, but from my reading it seems that multiprocesses side-step the GIL

auscompgeek: No. You would break the HAL, since you then have the HAL loaded in 2 processes.

Twinters007: Is that generally a bad thing? Why can you have it in two threads but not two processes?

virtuald: It was never built to be distributed across processes, so you can't.

Twinters007: Oh cause you're modifying data and such?

virtuald:
Worked on the roborio, so calling it good.

wpilib/wpilib/notifier.py
```diff
@@ -0,0 +1,124 @@
...
+        #: The thread waiting on the HAL alarm
+        self._thread = threading.Thread(target=self._thread, daemon=True)
+        self._thread.start()
+        
+        # python-specific
+        Resource._add_global_resource(self)
+    
+    # python-specific: this is called finalize in Java
+    def free(self) -> None:
+        handle, self._notifier = self._notifier, None
```
ariovistus: not that I can tell that whether it matters, but this is not an atomic operation

PR #387, opened 2017-12-26
Update drive related functions

Twinters007:


wpilib/wpilib/speedcontrollergroup.py
```diff
@@ -7,15 +7,35 @@
...
+        for speedController in args:
+            self.addChild(speedController)
         self.speedControllers = args
         self.isInverted = False
 
+        self.instances += 1
+        self.setName("SpeedControllerGroup", self.instances)
+
+
+
```
auscompgeek: 3 blank lines? That must be a record.

(You have a couple of 2 blank lines in other places, but I'm not going to spam you with review comments.)

Twinters007: Oh dustin is away isn't he...that means I can format my files

wpilib/wpilib/speedcontrollergroup.py
```diff
@@ -7,15 +7,33 @@
...
     def __init__(self, *args):
+        """Create a new SpeedControllerGroup with the provided SpeedControllers.
+
+        :param args: SpeedControllers to add
+        :type args: :class:`.SpeedController`
+        """
+        SendableBase.__init__(self)
+        SpeedController.__init__(self)
+
+        for speedController in args:
```
auscompgeek: Upstream requires at least one SpeedController to be passed. Shouldn't we check this? Could do it a similar way upstream does (i.e. specify one explicitly as an argument) - would be the simplest and most fool-proof way.

wpilib/wpilib/speedcontrollergroup.py
```diff
@@ -7,15 +7,33 @@
...
+            self.addChild(speedController)
         self.speedControllers = args
         self.isInverted = False
 
+        self.instances += 1
```
auscompgeek: I think this will always keep `SpeedControllerGroup.instances == 0`.

Twinters007: Do I have to do `SpeedControllerGroup.instances += 1`

auscompgeek: Yup. Same with the other ones.

wpilib/wpilib/speedcontrollergroup.py
```diff
@@ -26,9 +26,15 @@ def __init__(self, *args):
...
 
-        for speedController in args:
-            self.addChild(speedController)
-        self.speedControllers = args
+        self.speedControllers = []
+        self.speedControllers.append(speedController)
+        self.addChild(speedController)
+
+        for speedcontroller in args:
+            self.speedControllers.append(speedcontroller)
```
auscompgeek: tbh I would've preferred `self.speedControllers = [speedController] + args` but this works as well I guess

auscompgeek:
Might want to add some tests to make sure that those `instances` class variables actually increment.

wpilib/wpilib/speedcontrollergroup.py
Twinters007: This is closer to what they have.

Twinters007:
Just did

wpilib/wpilib/speedcontrollergroup.py
auscompgeek: True, but only because that's how you have to do it in Java. You might notice the C++ implementation is much more succinct.

auscompgeek: Should probably also change `args` to `speedControllers` too. Hmm.

auscompgeek:
race conditions/10

Appending an element you already know about to an empty list seems silly anyway. I think that change I suggested to SpeedControllerGroup would be much clearer.

wpilib/wpilib/speedcontrollergroup.py
```diff
@@ -26,15 +26,12 @@ def __init__(self, speedController, *args):
         SendableBase.__init__(self)
         SpeedController.__init__(self)
 
-        self.speedControllers = []
-        self.speedControllers.append(speedController)
+        self.speedControllers = [speedController] + list(args)
         self.addChild(speedController)
 
         for speedcontroller in args:
```
auscompgeek: Change this to `for speedcontroller in self.speedControllers` and you won't need the extra `self.addChild()` above.

Twinters007: o.O

PR #388, opened 2017-12-26
Fix encoder fpga bug

Twinters007:


Twinters007:
@auscompgeek if you merge this then @ariovistus 's test will pass and everything will be gucci

auscompgeek:
uhm no what don't you want to return `encoderHandle.idx`?

Twinters007:
I don't think the idx technically corresponds to the fpga index? It doesn't in the hal cpp implementation.

Twinters007:
Nvm some deep searching proved that it does get changed to the idx and whatnot

PR #389, opened 2017-12-26
HAL + SPI updates

virtuald:
I tested these on a RoboRIO before I left, so if the tests pass then I think they're fine unless there's something I accidentally included.

### DIFF
--- a/wpilib/wpilib/accumulatorresult.py
+++ b/wpilib/wpilib/accumulatorresult.py
@@ -1,0 +1,16 @@
+#----------------------------------------------------------------------------
+# Copyright (c) FIRST 2008-2017. All Rights Reserved.
+# Open Source Software - may be modified and shared by FRC teams. The code
+# must be accompanied by the FIRST BSD license file in the root directory of
+# the project.
+#----------------------------------------------------------------------------
+
+from collections import namedtuple
+
+__all__ = ['AccumulatorResult']
+
+AccumulatorResult = namedtuple('AccumulatorResult', ['value', 'count'])
+
+AccumulatorResult.__doc__ = 'Structure for holding the values stored in an accumulator.'
+AccumulatorResult.value.__doc__ = 'The total value accumulated.'
+AccumulatorResult.count.__doc__ = 'The number of sample value was accumulated over.'
--- a/wpilib/wpilib/spi.py
+++ b/wpilib/wpilib/spi.py
@@ -1,8 +1,16 @@
-# validated: 2016-12-20 JW b51e85ae262b edu/wpi/first/wpilibj/SPI.java
+# validated: 2016-12-20 JW e1195e8b9dab602f036da0c801ba0cc505730b80 edu/wpi/first/wpilibj/SPI.java
 
+import ctypes
+import enum
 import hal
+import struct
+import threading
 import warnings
 import weakref
+
+from .accumulatorresult import AccumulatorResult
+from .digitalsource import DigitalSource
+from .notifier import Notifier
 
 __all__ = ["SPI"]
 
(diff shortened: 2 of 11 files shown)


wpilib/wpilib/accumulatorresult.py
```diff
@@ -0,0 +1,16 @@
...
+# Open Source Software - may be modified and shared by FRC teams. The code
+# must be accompanied by the FIRST BSD license file in the root directory of
+# the project.
+#----------------------------------------------------------------------------
+
+from collections import namedtuple
+
+__all__ = ['AccumulatorResult']
+
+AccumulatorResult = namedtuple('AccumulatorResult', ['value', 'count'])
```
auscompgeek: Hm. Would it make sense to type hint this (using `typing.NamedTuple`)?

wpilib/wpilib/spi.py
```diff
@@ -198,103 +201,331 @@ def read(self, initiate, size):
...
+
+        kAccumulateDepth = 2048
+
+        def __init__(self, port: int, xferSize: int, validMask: int,
+                           validValue: int, dataShift: int, dataSize: int,
+                           isSigned: bool, bigEndian: bool):
+
+            # python-specific: for efficiency purposes, we only support xferSize of
+            #                  2, 4, or 8... if you need to do a different size then
+            #                  file a bug and we'll adjust it
```
auscompgeek: Hm. Is using `struct` actually any faster than `int.from_bytes`? I mean I wouldn't be surprised if it is, but I guess numbers would be nice.

wpilib/wpilib/spi.py
```diff
@@ -198,103 +201,331 @@ def read(self, initiate, size):
...
+            elif self._xferSize == 4:
+                self._struct = struct.Struct('%sI' % efmt)
+            elif self._xferSize == 8:
+                self._struct = struct.Struct('%sQ' % efmt)
+            else:
+                raise ValueError("SPI Accumulator only supports xferSize of 2/4/8")
+
+            self._mutex = threading.RLock()
+            self._notifier = Notifier(self._update)
+            self._buf = (ctypes.c_uint8 * (xferSize * self.kAccumulateDepth))()
```
auscompgeek: Would using a `bytearray` here work as well?

wpilib/wpilib/spi.py
```diff
@@ -198,103 +201,331 @@ def read(self, initiate, size):
...
-        hal.initSPIAccumulator(self.port, int(period*1.0e6), cmd,
-                               xfer_size, valid_mask, valid_value, data_shift,
-                               data_size, is_signed, big_endian)
+        self.initAuto(xferSize * 2048)
 
-    def freeAccumulator(self):
+        if bigEndian:
+            cmdBytes = cmd.to_bytes(4, 'big')
+        else:
+            cmdBytes = cmd.to_bytes(4, 'little')
```
auscompgeek: Tempting to rewrite this as a ternary expression.

wpilib/wpilib/accumulatorresult.py
virtuald: Knock yourself out.

wpilib/wpilib/spi.py
virtuald: I don't know.

wpilib/wpilib/spi.py
virtuald: well, from_bytes doesn't support a buffer offset? I think.

auscompgeek: Ah, good point; I missed that part. You could definitely use a `memoryview` to do that though.

auscompgeek:
Say, was spi.py missing an updated git-source-track line?

PR #391, opened 2017-12-26
verify AnalogPotentiometer some more

ariovistus:


Twinters007:
My source-track shows a diff in `get()`

ariovistus:
we seem to be calling the hal function directly. java seems to be calling an intermediate that calls the function.

Twinters007:
Yeah I guess that's cause they have to go through the JNI

PR #392, opened 2017-12-26
Update motorsafety and add tests

twinters007:


auscompgeek:
I don't see any tests failing without this alleged deadlock fix.

Twinters007:
Well the timedrobot example wouldn't let me set a motor value without it

auscompgeek:
You missed my point. Write a test for this.

Twinters007:
Dumb me didn't realize that the lock was an instance variable

auscompgeek:
Ok, this looks sane. Tests?

PR #394, opened 2017-12-27
Update livewindow changes

twinters007:
Didn't touch i2c because it's too far from wpilibj for me
to do comfortably

### DIFF
--- a/wpilib/tests/test_builtinaccelerometer.py
+++ b/wpilib/tests/test_builtinaccelerometer.py
@@ -36,24 +36,15 @@ def test_bacc_getZ(acc, acc_data):
     acc_data['z'] = 3.14
     assert acc.getZ() == 3.14
 
-def test_bacc_getSmartDashboardType(acc):
-    assert acc.getSmartDashboardType() == "3AxisAccelerometer"
-    
-def test_bacc_updateTable(acc, acc_data):
-    acc.xEntry = MagicMock()
-    acc.yEntry = MagicMock()
-    acc.zEntry = MagicMock()
+def test_bacc_initSendable(acc, sendablebuilder, acc_data):
     acc_data['x'] = 1
     acc_data['y'] = 2
     acc_data['z'] = 3
-    acc.updateTable()
 
-    acc.xEntry.setDouble.assert_has_calls([call(1)])
-    acc.yEntry.setDouble.assert_has_calls([call(2)])
-    acc.zEntry.setDouble.assert_has_calls([call(3)])
+    acc.initSendable(sendablebuilder)
+    sendablebuilder.updateTable()
 
-def test_bacc_updateTable_uninitialized(acc, acc_data):
-    acc.xEntry = None
-    acc.yEntry = None
-    acc.zEntry = None
-    acc.updateTable()
+    assert sendablebuilder.getTable().getValue("X", 0.0) == 1
+    assert sendablebuilder.getTable().getValue("Y", 0.0) == 2
+    assert sendablebuilder.getTable().getValue("Z", 0.0) == 3
+
--- a/wpilib/wpilib/adxl345_i2c.py
+++ b/wpilib/wpilib/adxl345_i2c.py
@@ -1,4 +1,4 @@
-# validated: 2017-09-20 AA 34c18ef00062 edu/wpi/first/wpilibj/ADXL345_I2C.java
+# validated: 2017-12-27 TW f9bece2ffbf7 edu/wpi/first/wpilibj/ADXL345_I2C.java
 #----------------------------------------------------------------------------
 # Copyright (c) FIRST 2008-2012. All Rights Reserved.
 # Open Source Software - may be modified and shared by FRC teams. The code
@@ -11,11 +11,11 @@ import hal
 from .interfaces import Accelerometer
 from .i2c import I2C
 from .sensorbase import SensorBase
-from .livewindow import LiveWindow
+from .sendable import Sendable
 
 __all__ = ["ADXL345_I2C"]
 
-class ADXL345_I2C(SensorBase):
+class ADXL345_I2C(SensorBase, Sendable):
     """
         ADXL345 accelerometer device via i2c
     """
@@ -66,10 +66,9 @@ class ADXL345_I2C(SensorBase):
         hal.report(hal.UsageReporting.kResourceType_ADXL345,
                       hal.UsageReporting.kADXL345_I2C)
 
-        LiveWindow.addSensor("ADXL345_I2C", port, self)
+        self.setName("ADXL345_I2C", port)
 
     def free(self):
-        LiveWindow.removeComponent(self)
         self.i2c.free()
         super().free()
 
@@ -145,27 +144,17 @@ class ADXL345_I2C(SensorBase):
                 rawData[1] * self.kGsPerLSB,
                 rawData[2] * self.kGsPerLSB)
 
-    # Live Window code, only does anything if live window is activated.
 
-    def getSmartDashboardType(self):
-        return "3AxisAccelerometer"
+    def _updateValues(self, entryX, entryY, entryZ):
+        data = self.getAccelerations()
+        entryX.setDouble(data[0])
+        entryY.setDouble(data[1])
+        entryZ.setDouble(data[2])
 
-    def initTable(self, subtable):
-        if subtable is not None:
-            self.xEntry = subtable.getEntry("X")
-            self.yEntry = subtable.getEntry("Y")
-            self.zEntry = subtable.getEntry("Z")
-            self.updateTable()
-        else:
-            self.xEntry = None
-            self.yEntry = None
-            self.zEntry = None
+    def initSendable(self, builder):
+        builder.setSmartDashboardType("3AxisAccelerometer")
+        entryX = builder.getEntry("X")
+        entryY = builder.getEntry("Y")
+        entryZ = builder.getEntry("Z")
 
-    def updateTable(self):
-        if self.xEntry is not None:
-            self.xEntry.setDouble(self.getX())
-        if self.yEntry is not None:
-            self.yEntry.setDouble(self.getY())
-        if self.zEntry is not None:
-            self.zEntry.setDouble(self.getZ())
-
+        builder.setUpdateTable(self._updateValues(entryX, entryY, entryZ))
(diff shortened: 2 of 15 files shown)


wpilib/wpilib/buttons/trigger.py
```diff
@@ -7,11 +7,15 @@
 #----------------------------------------------------------------------------
 
 from networktables.networktable import NetworkTable
+from wpilib import SendableBuilder
```
auscompgeek: uh.

wpilib/wpilib/analogtrigger.py
```diff
@@ -150,3 +158,7 @@ def createOutput(self, type):
         :returns: An AnalogTriggerOutput object.
         """
         return AnalogTriggerOutput(self, type)
+
+    def initSendable(self, builder):
+        if self.ownsAnalog:
+            self.analogInput.initSendable( )
```
auscompgeek: I feel there's something missing here.

wpilib/wpilib/analogaccelerometer.py
```diff
@@ -35,19 +34,27 @@ def __init__(self, channel):
...
+            self.allocatedChannel = False
+            self.analogChannel = channel
         self.voltsPerG = 1.0
         self.zeroGVoltage = 2.5
         self.pidSource = self.PIDSourceType.kDisplacement
         hal.report(hal.UsageReporting.kResourceType_Accelerometer,
                       self.analogChannel.getChannel())
-        LiveWindow.addSensorChannel("Accelerometer",
+        self.setName("Accelerometer",
                                     self.analogChannel.getChannel(), self)
```
auscompgeek: Something seems a little off.

I guess this means we have absolutely zero coverage of AnalogAccelerometer. Could you add a few basic tests?

wpilib/wpilib/sendablebuilder.py
```diff
@@ -146,7 +146,6 @@ def addBooleanProperty(self, key, getter, setter):
         :param getter:  getter function (returns current value)
         :type getter: () -> bool
         :param setter:  setter function (sets new value)
-        :type setter: (bool) -> Any
```
ariovistus: y?

wpilib/wpilib/builtinaccelerometer.py
```diff
@@ -9,11 +9,11 @@
 
 from .interfaces import Accelerometer
 from .sensorbase import SensorBase
-from .livewindow import LiveWindow
+from .sendable import Sendable
 
 __all__ = ["BuiltInAccelerometer"]
 
-class BuiltInAccelerometer(SensorBase):
+class BuiltInAccelerometer(SensorBase, Sendable):
```
ariovistus: Sendable is already in SensorBase's inheritance chain

wpilib/wpilib/analogtriggeroutput.py
```diff
@@ -45,6 +45,12 @@ class AnalogTriggerOutput(DigitalSource):
     Using the averaging engine may help with this, but rotational speeds of the
     sensor will then be limited.
     """
+    class AnalogTriggerType:
+        """Defines the state in which the :class:`.AnalogTrigger` triggers"""
```
ariovistus: is the only point of this class decl to insert a docstring?

wpilib/tests/test_builtinaccelerometer.py
```diff
@@ -36,24 +36,15 @@ def test_bacc_getZ(acc, acc_data):
...
     assert acc.getZ() == 3.14
 
-def test_bacc_getSmartDashboardType(acc):
-    assert acc.getSmartDashboardType() == "3AxisAccelerometer"
-    
-def test_bacc_updateTable(acc, acc_data):
-    acc.xEntry = MagicMock()
-    acc.yEntry = MagicMock()
-    acc.zEntry = MagicMock()
+def test_bacc_initSendable(acc, sendablebuilder, acc_data):
```
ariovistus: this test is basically suitable for all of the accelerometers

wpilib/wpilib/analogtriggeroutput.py
twinters007: No it gets used in other classes

wpilib/wpilib/sendablebuilder.py
twinters007: Cause it doesn't necessarily take a bool...although I thought I replaced it with...something.

wpilib/wpilib/analogtrigger.py
twinters007: Looks like I tried to do ^Space, ended up doing just space

wpilib/wpilib/analogtriggeroutput.py
ariovistus: seems elsewhere I've seen eg
```
AnalogTriggerType = hal.AnalogTriggerType
```

wpilib/wpilib/analogtrigger.py
```diff
@@ -67,10 +74,11 @@ def port(self):
 
     def free(self):
         """Release the resources used by this object"""
+        super().free()
         self.__finalizer()
         if self.analogInput:
             self.analogInput.free()
-            
+
```
ariovistus: 2 newlines?

wpilib/wpilib/digitalglitchfilter.py
```diff
@@ -12,6 +12,7 @@
 from .encoder import Encoder
 from .counter import Counter
 from .sensorbase import SensorBase
+from .sendable import Sendable
```
twinters007: I guess I don't need to import this...

wpilib/wpilib/adxl345_i2c.py
```diff
@@ -145,27 +144,17 @@ def getAccelerations(self):
...
-            self.yEntry.setDouble(self.getY())
-        if self.zEntry is not None:
-            self.zEntry.setDouble(self.getZ())
+    def initSendable(self, builder):
+        builder.setSmartDashboardType("3AxisAccelerometer")
+        entryX = builder.getEntry("X")
+        entryY = builder.getEntry("Y")
+        entryZ = builder.getEntry("Z")
 
+        builder.setUpdateTable(self._updateValues(entryX, entryY, entryZ))
```
auscompgeek: This doesn't seem like what you want. This would call `_updateValues`, which would then return None, so you're calling `builder.setUpdateTable(None)`. Did you mean to put that in a lambda?

Do we have a test for this? It seems odd to me that this would pass a decent test; if we do have a test already, maybe we need a more general and thorough test.

wpilib/wpilib/adxl345_spi.py
```diff
@@ -159,25 +158,16 @@ def getAccelerations(self):
...
-            self.yEntry.setDouble(self.getY())
-        if self.zEntry is not None:
-            self.zEntry.setDouble(self.getZ())
+    def initSendable(self, builder):
+        builder.setSmartDashboardType("3AxisAccelerometer")
+        entryX = builder.getEntry("X")
+        entryY = builder.getEntry("Y")
+        entryZ = builder.getEntry("Z")
 
+        builder.setUpdateTable(self._updateValues(entryX, entryY, entryZ))
```
auscompgeek: Ditto.

wpilib/wpilib/adxl362.py
```diff
@@ -179,27 +182,16 @@ def getAccelerations(self):
...
-            self.yEntry.setDouble(self.getY())
-        if self.zEntry is not None:
-            self.zEntry.setDouble(self.getZ())
+    def initSendable(self, builder):
+        builder.setSmartDashboardType("3AxisAccelerometer")
+        self._entryX = builder.getEntry("X")
+        self._entryY = builder.getEntry("Y")
+        self._entryZ = builder.getEntry("Z")
 
+        builder.setUpdateTable(self._updateValues)
```
auscompgeek: Ditto.

twinters007:
Yea it might be a python only thing, since in java Sendable is an interface, but most classes extend SensorBase, implement Sendable

auscompgeek:
In WPILibJ, SensorBase is an abstract class that extends SendableBase, which in turn also implements Sendable.

wpilib/wpilib/adxl362.py
auscompgeek: Rereading this again, I just realised this is correct. This is probably the simplest way to fix the above two.

wpilib/wpilib/analogtriggeroutput.py
twinters007: look at `counter.py`

wpilib/wpilib/buttons/trigger.py
```diff
@@ -7,11 +7,15 @@
 #----------------------------------------------------------------------------
 
 from networktables.networktable import NetworkTable
+from ..sendablebuilder import SendableBuilder
+
```
auscompgeek: The fact that a blank line separates these imports seems rather odd.

wpilib/wpilib/buttons/trigger.py
```diff
@@ -133,15 +136,13 @@ def execute():
...
-            self.pressedEntry.setBoolean(self.get())
-        else:
-            self.pressedEntry = None
+    def _safeState(self) -> None:
+        self.sendablePressed = False
+
+    def _setPressed(self, value: bool) -> None:
+        self.sendablePressed = value
+
+    def initSendable(self, builder: SendableBuilder):
```
auscompgeek: I actually don't think it's worth the trouble to type hint this method. We don't expect teams to call `initSendable` directly, so we don't actually need to document it. (sphinx-autodoc-typehints will create a docstring on type hinted functions if there isn't one already.)

wpilib/wpilib/sendable.py
```diff
@@ -54,20 +57,11 @@ def setSubsystem(self, subsystem):
         """
         raise NotImplementedError
 
-    def initSendable(self, builder):
+    def initSendable(self, builder: SendableBuilder) -> None:
         """
         Initializes this Sendable object.
 
         :param builder: sendable builder
         :type builder: :class:`wpilib.SendableBuilder`
```
auscompgeek: So this doesn't actually link right now. Given you're adding a type hint, perhaps remove this redundant line.

wpilib/wpilib/buttons/trigger.py
```diff
@@ -79,7 +82,7 @@ def execute():
         from ..command import Scheduler
```
auscompgeek: Heh. You didn't type hint the above method. Not that it really matters.

wpilib/wpilib/analogtriggeroutput.py
auscompgeek: I personally have no idea why it'd be here rather than AnalogTrigger, but guess this is the status quo.

PR #395, opened 2017-12-31
Remove SensorBase.free() method

auscompgeek:
Let these calls go up to SendableBase.free().

Upstream doesn't have a free method here anyway.


wpilib/wpilib/sensorbase.py
```diff
@@ -162,8 +162,3 @@ def getDefaultSolenoidModule():
         :returns: The number of the default solenoid module.
         """
         return SensorBase.defaultSolenoidModule
-
-    def free(self):
-        """Free the resources used by this object"""
-        # TODO: delete?
-        pass
```
twinters007: lol we're sending them up just to have nothing happen

PR #396, opened 2017-12-31
Fix WPILib spec scanner

auscompgeek:
* Update the WPILibJ tree path.
* Use `inspect.getfullargspec()` instead of the deprecated `inspect.getargspec()`, which cannot deal with Python 3 syntax.


twinters007:
Does it work?

auscompgeek:
Well, my intent was to fix this so I could use it, and I just used it, so I would say yes, it works with this.

PR #398, opened 2017-12-31
Count sim timer down rather than up

ErikBoesen:
`getMatchTime()` should return the number of seconds remaining in the simulated match, not the number of seconds for which the match has been running.

### DIFF
--- a/hal-sim/hal_impl/functions.py
+++ b/hal-sim/hal_impl/functions.py
@@ -103,16 +103,16 @@ def _initport(name, idx, status, root=hal_data):
     try:
         if idx < 0:
             raise IndexError()
-        
+
         data = root[name][idx]
     except IndexError:
         status.value = PARAMETER_OUT_OF_RANGE
         return
-        
+
     if data['initialized']:
         status.value = RESOURCE_IS_ALLOCATED
         return
-    
+
     data['initialized'] = True
     status.value = 0
     return data
@@ -209,10 +209,10 @@ def _getErrorMessage(code):
     elif code == HAL_PWM_SCALE_ERROR:
         return "HAL: The PWM Scale Factors are out of range"
     elif code == HAL_HANDLE_ERROR:
-        return  "HAL: A handle parameter was passed incorrectly"
+        return "HAL: A handle parameter was passed incorrectly"
     else:
         return "Unknown error status %s" % code
-    
+
 def getErrorMessage(code):
     return bytes(_getErrorMessage(code), 'utf-8')
 
@@ -238,7 +238,7 @@ def getFPGAButton(status):
 def getSystemActive(status):
     status.value = 0
     return True
-    
+
 def getBrownedOut(status):
     status.value = 0
     return False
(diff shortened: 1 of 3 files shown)


hal-sim/hal_impl/functions.py
```diff
@@ -1056,7 +1056,7 @@ def getMatchTime(status):
     status.value = 0
     match_start = hal_data['time']['match_start']
     if match_start is None:
-        return 0.0
+        return 135.0
     else:
         return (hooks.getFPGATime() - hal_data['time']['match_start'])/1000000.0
```
twinters007: This should be swapped.

ErikBoesen:
I've resolved the issues pointed out by @twinters007.

hal-sim/hal_impl/mode_helpers.py
```diff
@@ -5,114 +5,114 @@
...
         'autonomous': False,
         'test': False,
         'enabled': enabled,
         'dsAttached': True
     })
-    
+
     if enabled:
-        hal_data['time']['match_start'] = fns.hooks.getFPGATime() - 15000000
+        hal_data['time']['match_start'] = 135000000 - 15000000
```
twinters007: I think we should change the name of this key to `time_remaining` and currently, if I start the sim, wait 20 seconds, then enable again, the time won't be at the beginning of teleop, instead it'll say there are 100 seconds left. I don't think this is desired behavior. I think you should do `fns.hooks.getFPGATime() + 120000000`

ErikBoesen: I feel like just calling it `remaining` may be preferable, given that `time` is used in the parent dictionary.

ErikBoesen:
Fixed, with the caveat of calling the value `remaining` instead.

hal-sim/hal_impl/functions.py
```diff
@@ -1054,11 +1054,11 @@ def getMatchTime(status):
...
     status.value = 0
-    match_start = hal_data['time']['match_start']
-    if match_start is None:
-        return 0.0
+    remaining = hal_data['time']['remaining']
+    if remaining is None:
+        return 135.0
     else:
-        return (hooks.getFPGATime() - hal_data['time']['match_start'])/1000000.0
+        return (hal_data['time']['remaining'] - hooks.getFPGATime())/1000000.0
```
twinters007: Why do the dict lookup twice?

hal-sim/hal_impl/mode_helpers.py
```diff
@@ -5,114 +5,114 @@
...
         'autonomous': False,
         'test': False,
         'enabled': enabled,
         'dsAttached': True
     })
-    
+
     if enabled:
-        hal_data['time']['match_start'] = fns.hooks.getFPGATime() - 15000000
+        hal_data['time']['remaining'] = hal_data['time']['remaining'] + 12000000
```
twinters007: missing a 0. And if I enable autonomous (remaining = 135) then enable teleop, remaining now equals 255. Add to getFPGATime instead

PR #399, opened 2018-01-01
fix for notifier

ariovistus:


ariovistus:
it fixes the lack of the underscore

auscompgeek:
wow travis took an absurdly long amount of time to install deps in the 3.5 job here

PR #402, opened 2018-01-01
update smartdashboard for 2018 even more

ariovistus:


wpilib/wpilib/smartdashboard.py
```diff
@@ -516,3 +516,9 @@ def getRaw(cls, key, defaultValue):
         """
         table = cls.getTable()
         return table.getRaw(key, defaultValue)
+
+    @classmethod
+    def updateValues():
```
twinters007: missing a cls

auscompgeek:
Could you please also make the corresponding change in IterativeRobotBase?

auscompgeek:
I think you forgot to update the git-source-track line in IterativeRobotBase.

PR #406, opened 2018-01-04
Update wpilib.drive and fix docs

auscompgeek:
* Fix the broken drive base diagrams.
* Put the RobotDrive porting guides in note directives for clarity.
* Add the missing code snippets in DifferentialDrive.
* [upstream change] Invert the right motor on LiveWindow in DifferentialDrive.

### DIFF
--- a/wpilib/wpilib/drive/differentialdrive.py
+++ b/wpilib/wpilib/drive/differentialdrive.py
@@ -1,4 +1,4 @@
-# validated: 2018-01-04 DV bb38ef564204 edu/wpi/first/wpilibj/drive/DifferentialDrive.java
+# validated: 2018-01-04 DV 1f4822f33278 edu/wpi/first/wpilibj/drive/DifferentialDrive.java
 # ----------------------------------------------------------------------------
 # Copyright (c) FIRST 2008-2018. All Rights Reserved.
 # Open Source Software - may be modified and shared by FRC teams. The code
@@ -28,12 +28,12 @@ class DifferentialDrive(RobotDriveBase):
     Four motor drivetrain::
 
         def robotInit(self):
-            self.frontLeft = wpilib.Talon(1)
-            self.rearLeft = wpilib.Talon(2)
+            self.frontLeft = wpilib.Spark(1)
+            self.rearLeft = wpilib.Spark(2)
             self.left = wpilib.SpeedControllerGroup(self.frontLeft, self.rearLeft)
 
-            self.frontRight = wpilib.Talon(3)
-            self.rearRight = wpilib.Talon(4)
+            self.frontRight = wpilib.Spark(3)
+            self.rearRight = wpilib.Spark(4)
             self.right = wpilib.SpeedControllerGroup(self.frontRight, self.rearRight)
 
             self.drive = DifferentialDrive(self.left, self.right)
@@ -41,14 +41,14 @@ class DifferentialDrive(RobotDriveBase):
     Six motor drivetrain::
 
         def robotInit(self):
-            self.frontLeft = wpilib.Talon(1)
-            self.midLeft = wpilib.Talon(2)
-            self.rearLeft = wpilib.Talon(3)
+            self.frontLeft = wpilib.Spark(1)
+            self.midLeft = wpilib.Spark(2)
+            self.rearLeft = wpilib.Spark(3)
             self.left = wpilib.SpeedControllerGroup(self.frontLeft, self.midLeft, self.rearLeft)
 
-            self.frontRight = wpilib.Talon(4)
-            self.midRight = wpilib.Talon(5)
-            self.rearRight = wpilib.Talon(6)
+            self.frontRight = wpilib.Spark(4)
+            self.midRight = wpilib.Spark(5)
+            self.rearRight = wpilib.Spark(6)
             self.right = wpilib.SpeedControllerGroup(self.frontRight, self.midRight, self.rearRight)
 
             self.drive = DifferentialDrive(self.left, self.right)
--- a/wpilib/wpilib/drive/differentialdrive.py
+++ b/wpilib/wpilib/drive/differentialdrive.py
@@ -22,6 +22,37 @@ class DifferentialDrive(RobotDriveBase):
     These drive bases typically have drop-center / skid-steer with two or more wheels per side
     (e.g., 6WD or 8WD).
 
+    This class takes a :class:`.SpeedController` per side. For four and six motor
+    drivetrains, construct and pass in :class:`.SpeedControllerGroup` instances as follows.
+
+    Four motor drivetrain::
+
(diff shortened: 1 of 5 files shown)


wpilib/wpilib/drive/differentialdrive.py
```diff
@@ -70,13 +101,13 @@ class DifferentialDrive(RobotDriveBase):
     def __init__(self, leftMotor, rightMotor):
         """Constructor for DifferentialDrive.
 
-        .. note:: This class takes in a SpeedController per side. For two and four motor drivetrains,
-                  construct and pass in  :class:`.SpeedControllerGroup` instances as follows.
+        .. note:: To pass multiple motors per side, use a :class:`.SpeedControllerGroup`.
```
twinters007: While your docs are better than upstreams, I think we should keep upstreams

wpilib/wpilib/drive/differentialdrive.py
```diff
@@ -70,13 +101,13 @@ class DifferentialDrive(RobotDriveBase):
...
+                  If a motor needs to be inverted, do so before passing it in.
 
-        :param leftMotor: Left motor SpeedController
-        :param rightMotor: Right motor SpeedController
-        :type leftMotor: :class:`.SpeedController` or :class:`.SpeedControllerGroup`
-        :type rightMotor: :class:`.SpeedController` or :class:`.SpeedControllerGroup`
+        :param leftMotor: Left motor(s)
+        :param rightMotor: Right motor(s)
+        :type leftMotor: :class:`.SpeedController`
```
twinters007: Why get rid of the or? I guess a SpeedControllerGroup is a speed controller...But it makes it clearer in the docs imo

wpilib/wpilib/drive/differentialdrive.py
```diff
@@ -95,7 +126,6 @@ def __init__(self, leftMotor, rightMotor):
 
     def arcadeDrive(self, xSpeed, zRotation, squaredInputs=True):
         """Arcade drive method for differential drive platform.
-        The calculated values will be squared to decrease sensitivity at low speeds
```
twinters007: What are you doing?

wpilib/wpilib/drive/robotdrivebase.py
```diff
@@ -52,24 +52,23 @@ def __init__(self):
...
 
-    def setDeadband(self, deadband):
-        """Change the value for deadband scaling. The default value is `.02`. Values smaller than the
-         deadband are set to 0, while values larger than the deadband are scaled from 0.0 to 1.0. See
-         :meth:`applyDeadband()`
-
-         :param deadband: The deadband to set
-         :type deadband: float
-         """
+    def setDeadband(self, deadband: float):
```
twinters007: Why here and not everywhere else

wpilib/wpilib/drive/robotdrivebase.py
```diff
@@ -52,24 +52,23 @@ def __init__(self):
...
+        """
         self.deadband = deadband
 
-    def setMaxOutput(self, maxOutput):
+    def setMaxOutput(self, maxOutput: float):
         """Configure the scaling factor for using drive methods with motor controllers in a mode
         other than PercentVbus or to limit the maximum output.
 
-        The default value is `1.0`
+        The default value is :data:`kDefaultMaxOutput`.
```
twinters007: In my (limited) experience, this doesn't resolve and just shows "kDefaultMaxOutput". Am I wrong?

wpilib/wpilib/drive/killoughdrive.py
```diff
@@ -1,4 +1,11 @@
 # validated: 2017-12-25 TW f9bece2ffbf7 edu/wpi/first/wpilibj/drive/KilloughDrive.java
+#----------------------------------------------------------------------------
+# Copyright (c) FIRST 2017. All Rights Reserved.
```
twinters007: They've updated all of the copyrights to be x-2018. So this should be 2017-2018

wpilib/wpilib/drive/mecanumdrive.py
```diff
@@ -1,4 +1,11 @@
 # validated: 2017-12-25 TW f9bece2ffbf7 edu/wpi/first/wpilibj/drive/MecanumDrive.java
+#----------------------------------------------------------------------------
+# Copyright (c) FIRST 2008-2017. All Rights Reserved.
```
twinters007: Same thing

wpilib/wpilib/drive/differentialdrive.py
auscompgeek: This is what upstream has.

wpilib/wpilib/drive/differentialdrive.py
auscompgeek: This isn't accurate if you pass in `squaredInputs=False`.

wpilib/wpilib/drive/robotdrivebase.py
auscompgeek: You are indeed wrong. It links fine.

wpilib/wpilib/drive/robotdrivebase.py
auscompgeek: I don't see anywhere else in this file that has type information. Too lazy to add in the missing types.

wpilib/wpilib/drive/differentialdrive.py
auscompgeek: It also made the docs not that pretty. Given there's a note right above it, I don't think it's very useful nor practical to list it again here.

wpilib/wpilib/drive/differentialdrive.py
twinters007: Make a PR upstream?

wpilib/wpilib/drive/robotdrivebase.py
twinters007: Does it just link or will it replace it with '1.0'

auscompgeek: It doesn't show the value. Although I just realised the correct directive would be `:const:`, but not even that shows the value by default. We can easily show the value instead, but we'd have to remember to change it if it changes.

wpilib/wpilib/drive/differentialdrive.py
twinters007: Why can't I read?

wpilib/wpilib/drive/differentialdrive.py
auscompgeek: Upstream doesn't have this for `arcadeDrive(xSpeed, zRotation, squaredInputs)`.

twinters007: Oh it's just for the default...I see

PR #408, opened 2018-01-05
Deprecate ControllerPower

twinters007:


auscompgeek:
I think having a single deprecated message at the class level is enough. I think we should make the RobotController mention a link too.

If you think it'd be worth adding DeprecationWarnings, I could throw in some decorators, but I suspect this class is so disused that probably nobody would see them anyway.

PR #411, opened 2018-01-06
Bump pynetworktables dependency

auscompgeek:
Also reorder the requirements.txt so that pynetworktables is actually
installed first.  This eliminates the need for the ROBOTPY_NO_DEPS
environment variable for wpilib's setup.py.


twinters007:
🙏

PR #412, opened 2018-01-06
Update driverstation

twinters007:


wpilib/wpilib/driverstation.py
```diff
@@ -66,15 +77,22 @@ def __init__(self):
...
             raise ValueError("Do not create DriverStation instances, use DriverStation.getInstance() instead")
 
-        self.joystickMutex = threading.RLock()
+        self.cacheDataMutex = threading.RLock()
 
         self.joystickAxes = [hal.JoystickAxes() for _ in range(self.kJoystickPorts)]
         self.joystickPOVs = [hal.JoystickPOVs() for _ in range(self.kJoystickPorts)]
         self.joystickButtons = [hal.JoystickButtons() for _ in range(self.kJoystickPorts)]
+        self.matchInfo = self.MatchInfoData()
+        hal.getMatchInfo(self.matchInfo)
```
virtuald: This isn't going to work.

wpilib/wpilib/driverstation.py
```diff
@@ -33,6 +34,16 @@ class Alliance:
         Blue = 1
         Invalid = 2
 
+    class MatchType(enum.IntEnum):
+        None_ = 0
+        Practice = 1
+        Qualification = 2
+        Elimination = 3
+
+    class MatchInfoData:
```
virtuald: This is already defined in HAL, I think if you use that then it will work.

wpilib/wpilib/driverstation.py
auscompgeek: I'm pretty sure this is the wrong place to call this.

wpilib/wpilib/driverstation.py
```diff
@@ -125,23 +143,21 @@ def _reportErrorImpl(isError, code, error, printTrace):
...
-            locString = "%s.%s:%s" % (stack[-1][0], stack[-1][1], stack[-1][2])
+        exc = sys.exc_info()[0]
+        stack = traceback.extract_stack()[:-2]  # last one is this func
+        if exc is not None: # i.e. if an exception is present
+            # remove call of full_stack, the printed exception
+            # will contain the caught exception caller instead
+            del stack[-1]
+            del stack[-1]
 
+        locString = "%s.%s:%s" % (stack[-1][0], stack[-1][1], stack[-1][2])
```
auscompgeek: Why was this changed?

PR #416, opened 2018-01-09
Fix AnalogTrigger initialization

virtuald:


auscompgeek:
Do our tests not cover both branches here? 😕

virtuald:
At some point we should fix the sim to test for this, but I think this is fine for now.

virtuald:
I don't think the sim is implemented to check that it's getting an input rather than a port.

PR #420, opened 2018-01-10
Provide printTrace parameter in calls to DriverStation.reportWarning/Error

ErikBoesen:
Currently a new warning is caused whenever the `Robots shouldn't quit...` error is reported. In wpilibj the `printTrace` argument is `false` in all calls, so I've updated it to be that way here as well.

### DIFF
--- a/wpilib/wpilib/robotbase.py
+++ b/wpilib/wpilib/robotbase.py
@@ -16,14 +16,14 @@ __all__ = ["RobotBase"]
 
 class RobotBase:
     """Implement a Robot Program framework.
-    
+
     The RobotBase class is intended to be subclassed by a user creating a
     robot program.  Overridden ``autonomous()`` and ``operatorControl()`` methods
     are called at the appropriate time as the match proceeds. In the current
     implementation, the Autonomous code will run to completion before the
     OperatorControl code could start. In the future the Autonomous code might
     be spawned as a task, then killed at the end of the Autonomous period.
-    
+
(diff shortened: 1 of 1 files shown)


wpilib/wpilib/robotbase.py
```diff
@@ -173,12 +173,12 @@ def main(robot_cls):
         except:
             from .driverstation import DriverStation
             DriverStation.reportError("Unhandled exception", True)
-            DriverStation.reportWarning("Robots should not quit, but yours did!")
+            DriverStation.reportWarning("Robots should not quit, but yours did!", False)
             DriverStation.reportError("The startCompetition() method (or methods called by it) should have handled the exception above.")
```
twinters007: Can you fix all of them? This is wrong too. 6 total. You got 3

wpilib/wpilib/robotbase.py
```diff
@@ -147,15 +147,15 @@ def main(robot_cls):
         except:
             from .driverstation import DriverStation
             DriverStation.reportError("Unhandled exception instantiating robot " + robot_cls.__name__, True)
-            DriverStation.reportWarning("Robots should not quit, but yours did!")
+            DriverStation.reportWarning("Robots should not quit, but yours did!", False)
             DriverStation.reportError("Could not instantiate robot "+robot_cls.__name__+"!")
```
twinters007: Can you get all 6 of these? (3 you have, 3 more to go)

PR #421, opened 2018-01-10
Fix driverstation bug (419b)

twinters007:


wpilib/wpilib/driverstation.py
```diff
@@ -657,7 +657,7 @@ def _getData(self):
             hal.getJoystickPOVs(stick, self.joystickPOVsCache[stick])
             hal.getJoystickButtons(stick, self.joystickButtonsCache[stick])
 
-        hal.getMatchInfo(self.matchInfoCache)
+        self.matchInfoCahce = hal.MatchInfo()
```
virtuald: This is misspelled, and it's no longer calling getMatchInfo.

twinters007: https://github.com/robotpy/robotpy-wpilib/pull/414/files#diff-cffd12ee377c3edf8d638d6ca2cf25c5R87

I may have jumped the gun on functionality here. Should these not operate the same?

wpilib/wpilib/driverstation.py
```diff
@@ -657,7 +657,7 @@ def _getData(self):
             hal.getJoystickPOVs(stick, self.joystickPOVsCache[stick])
             hal.getJoystickButtons(stick, self.joystickButtonsCache[stick])
 
-        hal.getMatchInfo(self.matchInfoCache)
+        _, self.matchInfoCache = hal.getMatchInfo()
```
virtuald: Have you tried executing this on a rio? Pretty sure it doesn't work.

twinters007: No. No rio anymore. What makes you think it doesn't work?

virtuald: Looks like it only returns a single value: https://github.com/robotpy/robotpy-wpilib/blob/master/hal-base/hal/functions.py#L368

auscompgeek:
I'm not sure what bug you're trying to fix. Whatever it is, I highly doubt it's here.

wpilib/wpilib/driverstation.py
twinters007: https://github.com/robotpy/robotpy-wpilib/blob/master/hal-sim/hal_impl/functions.py#L1063

......o.O

twinters007:
It looks like something that should've been fixed in 419. 

https://pastebin.com/zHU8Ecyu

auscompgeek:
That paste is unreadable.

twinters007:
Yeah it's from the DS log. It's the best I got

PR #422, opened 2018-01-11
Fix _freeMatchInfo TypeError on roboRIO

auscompgeek:
This should fix that error that @twinters007 was encountering. Hard to tell when it's an unreadable copy-paste from the DS log though.

Closes: #421


auscompgeek:
Just tested this on a roboRIO, and it looks like it works fine.

virtuald:
If it works, great.

PR #424, opened 2018-01-12
Don't decode string gameSpecificMessage

ErikBoesen:
This was causing some issues for me, see screenshot. From my reading of the [Control System docs](http://wpilib.screenstepslive.com/s/currentCS/m/getting_started/l/826278-2018-game-data-details), the `gameSpecificMessage` is returned as a string.
gsm

### DIFF
--- a/hal-sim/hal_impl/functions.py
+++ b/hal-sim/hal_impl/functions.py
@@ -1067,7 +1067,7 @@ def getMatchInfo():
     info.matchType = evt['match_type']
     info.matchNumber = evt['match_number']
     info.replayNumber = evt['replay_number']
-    info.gameSpecificMessage = bytes(evt['game_specific_message'], 'utf-8')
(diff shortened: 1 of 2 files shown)


auscompgeek:
Uhh… is this in simulation, or on the roboRIO? What versions of wpilib and hal do you have installed? (These are printed when the robot code starts.)

There is absolutely no way that this error can occur if everything is up to date.

ErikBoesen:
Pretty sure I had 2018.0.4 on everything. I will make sure tomorrow when I have robot access again.

auscompgeek:
Well, I can't reproduce this in sim with v2018.0.4, and C strings definitely don't have the type `str` either.

ErikBoesen:
Hmm, I guess some version conflict then. Sorry to bother you!

@twinters007

PR #425, opened 2018-01-12
Update drive. Urgent. 

twinters007:
Right side motors weren't being reversed in arcade or curvature drive.

### DIFF
--- a/wpilib/tests/test_drive.py
+++ b/wpilib/tests/test_drive.py
@@ -123,16 +123,18 @@ def drive_mecanum(wpimock, halmock):
     m4.reset_mock()
     return drive
 
-
+quickStopAccumulator = 0
 def check_curvature(wpimock, drive_diff, y, rotation, isQuickTurn):
-    quickStopAccumulator = 0
+    global quickStopAccumulator
     y = wpimock.drive.RobotDriveBase.limit(y)
     y = wpimock.drive.RobotDriveBase.applyDeadband(y, drive_diff.deadband)
 
+    rotation = wpimock.drive.RobotDriveBase.limit(rotation)
+    rotation = wpimock.drive.RobotDriveBase.applyDeadband(rotation, drive_diff.deadband)
+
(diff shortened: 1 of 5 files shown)


wpilib/wpilib/robotdrive.py
```diff
@@ -29,7 +29,7 @@ def _freeRobotDrive(allocatedSpeedControllers):
 class RobotDrive(MotorSafety):
     """
     .. deprecated:: 2018.0.0
-        Use DifferentialDrive, KilloughDrive, or MecanumDrive instead.
+        Use :class:`.DifferentialDrive`, :class:`.KilloughDrive`, or :class:`.MecanumDrive` instead.
```
auscompgeek: KilloughDrive didn't replace anything in RobotDrive

wpilib/tests/test_drive.py
```diff
@@ -123,16 +123,18 @@ def drive_mecanum(wpimock, halmock):
     m4.reset_mock()
     return drive
 
-
+quickStopAccumulator = 0
 def check_curvature(wpimock, drive_diff, y, rotation, isQuickTurn):
-    quickStopAccumulator = 0
+    global quickStopAccumulator
```
auscompgeek: why is this a global

twinters007: I changed it to act as a real accumulator for the tests. Otherwise it would get overwritten every time the function changed. None of our test cases rely on this, but I think it'd be good to allow for it in the future.

auscompgeek: I don't think globals in tests work like that

twinters007: really?

twinters007: hm looks like you're right. I'd have to use a fixture

ErikBoesen:
Would be nice to get this merged...

twinters007:
@auscompgeek

PR #430, opened 2018-01-18
Fix LiveWindowSendable.setName

auscompgeek:


virtuald:
I'm a bit confused by the Sendable/LiveWindowSendable/SendableBase set of interfaces... I'm not sure why the underscore was chosen for one and not the other. Seems like we should use kw arguments instead of two functions?

auscompgeek:
I believe it's because only Java's `Sendable.setName(name)` is intended to be required to be implemented - the other setName signatures have default implementations.

auscompgeek:
Heh, derp. This is still actually broken, because the Sendable interface is just plain broken.

The easiest way to fix this would have probably been to reorder the arguments in `Sendable.setName`, but I think it's a bit too late for that now.

Changing LiveWindowSendable to subclass SendableBase would fix LiveWindowSendable, but we'd still have the issue with the Sendable class being broken.

auscompgeek:
We really should prefer keyword-only arguments instead of the `match_arglist` mess when we can.

PR #431, opened 2018-01-18
DriverStation: Check None-ness before decode

auscompgeek:
If the game specific message or the event name are queried before the
driver station connects to the roboRIO, these will be null.
In this scenario, we should return None as is rather than trying to
decode it (which would obviously fail with an AttributeError).

### DIFF
--- a/wpilib/wpilib/driverstation.py
+++ b/wpilib/wpilib/driverstation.py
@@ -499,7 +499,9 @@ class DriverStation:
         :returns: The game specific message
         """
         with self.cacheDataMutex:
-            return self.matchInfo.gameSpecificMessage.decode('utf-8')
+            message = self.matchInfo.gameSpecificMessage
+        if message is not None:
+            return message.decode('utf-8')
 
     def getEventName(self) -> str:
         """Get the event name
@@ -507,7 +509,9 @@ class DriverStation:
         :returns: The event name
         """
         with self.cacheDataMutex:
-            return self.matchInfo.eventName.decode('utf-8')
+            name = self.matchInfo.eventName
+        if name is not None:
+            return name.decode('utf-8')
 
     def getMatchType(self) -> MatchType:
         """Get the match type.


wpilib/wpilib/driverstation.py
```diff
@@ -499,15 +499,19 @@ def getGameSpecificMessage(self) -> str:
         :returns: The game specific message
         """
         with self.cacheDataMutex:
-            return self.matchInfo.gameSpecificMessage.decode('utf-8')
+            message = self.matchInfo.gameSpecificMessage
+        if message is not None:
```
virtuald: Hm, what does Java return in this case then? null or empty string? Should update the docstring to indicate that None can be returned.

auscompgeek: I _think_ both C++ and Java would return a null pointer if the DS hasn't connected yet, judging by the implementation. Unfortunately that doesn't appear to be documented anywhere, not even on [the Screensteps documentation](http://wpilib.screenstepslive.com/s/currentCS/m/getting_started/l/826278-2018-game-data-details).

I think it would be beneficial to ask… someone who has edit access to the Screensteps docs to clarify that over there.

virtuald: Looking at DriverStationData.cpp, it appears to always allocate a string, so this should never return None.

virtuald: But... 

```
>>> m = hal.MatchInfo()
>>> mm = hal.getMatchInfo(m)
>>> m
<hal_impl.types.MatchInfo object at 0xb6203a30>
>>> m.
m.eventName            m.gameSpecificMessage  m.matchNumber          m.matchType            m.replayNumber
>>> m.eventName is None
True
>>> m.gameSpecificMessage is None
True
>>> m.matchNumber
0
>>> m.matchType
0
>>> m.replayNumber
0
>>>
```

virtuald: Aha, it's because `_getMatchInfo` returns an error code, so the object isn't filled in. However, instead of None, you should return an empty string, because that's what the documentation implies (and I think that's what will happen in C++ and Java).

auscompgeek: I'm not sure the documentation does imply that. Either way, I wouldn't be confident in an implication like that.

DriverStationData.cpp is simulation only, so that's not really useful.

Poking around the hal package in WPILibJ, it definitely looks like eventName and gameSpecificMessage are initialised to null, rather than the empty string. (See MatchInfoData.java - we're considering the scenario where setData hasn't been called yet.)

virtuald:
Merging for now, and filed a bug at https://github.com/wpilibsuite/allwpilib/issues/895

PR #433, opened 2018-01-18
Point hal-roborio at release versions

auscompgeek:


virtuald:
Oh ha, I was wondering why the versions were weird numbers.

PR #442, opened 2018-02-02
Add tip about writeBulk() to transaction() docs

codetheweb:
Make differences between transaction() and writeBulk() more clear.

### DIFF
--- a/wpilib/wpilib/i2c.py
+++ b/wpilib/wpilib/i2c.py
@@ -82,7 +82,9 @@ class I2C:
         """Generic transaction.
 
         This is a lower-level interface to the I2C hardware giving you more
-        control over each transaction.
+        control over each transaction. If you intend to write multiple bytes
+        in the same transaction and do not plan to receive anything back, use
+        writeBulk() instead.
 
         :param dataToSend: Buffer of data to send as part of the transaction.
         :type dataToSend: iterable of bytes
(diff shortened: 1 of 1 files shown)


auscompgeek:
This should also definitely go [upstream](/wpilibsuite/allwpilib) too.

virtuald:
I think this should be pushed upstream as well. I think the wording isn't quite right though -- didn't your report say that if you are trying to read zero bytes, that this fails? That really should be what's documented -- and if it's not how it should work, then we should address it so that it does work.

codetheweb:
@auscompgeek add it [here](https://github.com/wpilibsuite/allwpilib/blob/5ca00dddbeffd825f41efa17259574e2d75f4d08/hal/src/main/native/athena/I2C.cpp#L92)?  Sorry, I'm not very familiar with WPILIB.

From some initial searching through Github Issues on wpilibsuite/allwpilib, it appears they're using NI provided libraries called `libi2c` and `libspi` for I2C and SPI communication.  It sounds like both libraries are fairly buggy and there is little to no documentation on them.

Since it's unclear whether this is the intended behavior (and the maintainers of wpilibsuite/allwpilib are unlikely to know), I'll just modify the wording.

auscompgeek:
I'm pretty sure libi2c and libspi are the open source Linux libraries for I2C and SPI, judging by the HAL source code.

You wouldn't really want to change the HAL source, since teams won't see any docs there. You'll want to add them to the WPILibC and WPILibJ docs.

codetheweb:
(See my edit, @auscompgeek is right.)

Opened a [pull request](https://github.com/wpilibsuite/allwpilib/pull/929) upstream.

Does this look good to merge?

PR #451, opened 2018-02-15
TimedRobot now uses the HAL Notifier API directly

virtuald:
- Less threads is a good thing
- See https://github.com/wpilibsuite/allwpilib/pull/942

Tried it on a RoboRIO and it works as expected. Depends on https://github.com/robotpy/pyfrc/issues/94 to be fixed for it to work in sim/test.

### DIFF
--- a/wpilib/wpilib/timedrobot.py
+++ b/wpilib/wpilib/timedrobot.py
@@ -10,6 +10,8 @@ import hal
 
 from .iterativerobotbase import IterativeRobotBase
 from .notifier import Notifier
+from .resource import Resource
+from .robotcontroller import RobotController
 from .timer import Timer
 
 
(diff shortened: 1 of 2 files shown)


wpilib/wpilib/timedrobot.py
```diff
@@ -33,7 +35,14 @@ def __init__(self):
...
         self.startLoop = False
-        self.loop = Notifier(self.loopFunc)
+        
+        self._expirationTime = 0
+        self._notifier = hal.initializeNotifier()
+        
+        Resource._add_global_resource(self)
+    
+    def free(self) -> None:
+        hal.stopNotifier(self._notifier)
```
auscompgeek: Wouldn't you want to call `hal.cleanNotifier`?

PR #452, opened 2018-02-15
Report cscore usage reporting data

virtuald:
- Not complete, but good enough for now I suppose
- Fixes https://github.com/robotpy/robotpy-cscore/issues/25

@PeterJohnson @auscompgeek thoughts? Seems good enough to me.


auscompgeek:
This was the basic Good Enough Fix I had in mind too. Seems reasonable enough.

PR #454, opened 2018-02-22
DriverStation: Send actual control word over NT

auscompgeek:
This allows Shuffleboard's basic FMS info widget to correctly report the robot status.

Whilst I haven't had a chance to test this on a roboRIO, the ctypes magic works correctly in simulation as well (it was actually the original simulation implementation too, but I decided against using in simulation).

I'm still unsure whether this is actually the approach we should take in simulation, or whether we should actually use the ctypes hackery in simulation too. (I think we do want to use the ctypes magic on the roboRIO though, as I believe it will be faster, but again I haven't checked that.)


virtuald:
Tested on a RoboRIO and simulator, looks good to me.

PR #463, opened 2018-03-08
Avoid adding to LiveWindow where WPILibJ doesn't

auscompgeek:
None of these classes actually call the SendableBase constructor
upstream, so none of these are actually automatically added to
LiveWindow.


virtuald:
... that seems unintentional

auscompgeek:
I too think this was unintentional

virtuald:
Referring to Austin's reply on https://github.com/wpilibsuite/allwpilib/issues/1031, if no super constructor is called, the no-arg constructor is automatically called. https://docs.oracle.com/javase/tutorial/java/IandI/super.html

PR #467, opened 2018-05-04
Clean up Javadoc converter

auscompgeek:
This cleans up the Javadoc converter a bit.

A Python port, for convenience (think `:r !sphinxify`), was originally a part of this PR, but has been moved to https://github.com/auscompgeek/sphinxify (`pip3 install sphinxify`).

### DIFF
--- a/devtools/convert_javadoc.html
+++ b/devtools/convert_javadoc.html
@@ -99,7 +99,11 @@
         
         text += '\r\n' + pname + ' '.repeat(pindent.length - pname.length) + pp[0] + "\r\n";
         for (j = 1; j < pp.length; j++) {
-          text += pindent + pp[j] + "\r\n";
+          if (pp[j]) {
+            text += pindent + pp[j] + "\r\n";
+          } else {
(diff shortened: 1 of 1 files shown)


devtools/convert_javadoc.html
```diff
@@ -234,23 +240,13 @@
...
-            break;
-          case "staticmethod":
-            if (args.startsWith(",")) {
-              args = args.substring(1);
-            }
-            t = "@staticmethod\n    def " + funcName + "(" + args +
-                  ") -> " + retType + ":\n        " + t;
-            break;
+
+        if (modifiers.includes("static")) {
```
virtuald: Why did you get rid of staticmethod?

auscompgeek: Because (Guido admitted) staticmethod should never have existed in Python in the first place.

auscompgeek:
Splitting out the Python version into a package is probably a great idea. I'll probably create a PyPI package tomorrow morning. I think I'll call it sphinxize.

PR #468, opened 2018-05-05
[WIP] Lazy import the wpilib package

auscompgeek:
This is an attempt at lazily loading the wpilib package.

Currently only tested on CPython 3.6.

~~Still need to fix docs generation and the documentation tests…~~


twinters007:
Travis will test on 3.4 and 3.5. What does the `time` command output for importing now?

wpilib/wpilib/__init__.py
```diff
@@ -2,87 +2,44 @@
...
+        self._impl = _impl  # noqa: F821
+        self.run = run
+
+        # Re-export things used by the Python import machinery.
+        self.__name__ = __name__
+        self.__package__ = __package__
+        self.__path__ = __path__
+
+    def __getattr__(self, name):
+        if not name[0].isupper():
```
twinters007: Remind me, are drives imported with `drive.XXXDrive`?

auscompgeek: One would need to `import wpilib.drive` first.

twinters007: and wouldn't that break this?

twinters007: and wouldn't that break this?

auscompgeek:
Well, this PR would mean that timing only an import of the wpilib package would be a near-useless metric now. A much more useful metric would be "how long does it take to go from interpreter startup to disabledInit?"

auscompgeek:
Also Travis doesn't test on PyPy3 (where things currently still work, I believe). Might be worth adding pypy3 to the Travis config…

wpilib/wpilib/__init__.py
auscompgeek: No? Why would it?

wpilib/wpilib/__init__.py
```diff
@@ -2,87 +2,44 @@
...
+            raise AttributeError("module 'wpilib' has no attribute {!r}".format(name))
+        mod = importlib.import_module('.' + name.lower(), 'wpilib')
+        cls = getattr(mod, name)
+        setattr(self, name, cls)
+        return cls
+
+
+sys.modules['wpilib'] = WPILibLazyLoader()
```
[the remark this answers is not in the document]
auscompgeek: Uh, I guess if we put in a `__all__` it won't break IDEs. I guess that'd be useful for the docs generation too.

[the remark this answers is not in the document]
auscompgeek: Ugh, jedi doesn't listen to `__all__`.  That's annoying.  I guess this really is going to screw over IDEs.

auscompgeek:
Uhm. I don't know why the tests are failing. Not a clue. The more I stare into the abyss, the more I think this is not a good idea.

auscompgeek:
This experiment has been a wild ride, but I don't think this is worth it at this point. We can live for now.

virtuald:
I'm guessing the tests are failing because networktables isn't initialized and neither is hal (presumably because imports weren't done).

wpilib/wpilib/__init__.py
twinters007: cause if the first letter of the thing being imported isn't uppercase it'll throw an error?

auscompgeek: Importing a module within a package sets the attribute on the package.

PR #470, opened 2018-05-23
Add type hinting

ArchdukeTim:


wpilib/wpilib/adxl345_i2c.py
```diff
@@ -52,6 +53,7 @@ def __init__(self, port, range, address=None):
         :type range: :class:`.ADXL345_I2C.Range`
         :param address: the I2C address of the accelerometer (0x1D or 0x53)
         """
+        super().__init__()
```
virtuald: I hate to be pedantic... but this has nothing to do with type hinting.

wpilib/wpilib/adxl362.py
```diff
@@ -43,19 +46,20 @@ class ADXL362(SensorBase):
...
-    def __init__(self, range, port=None):
+    def __init__(self, range: Range, port=None):
         """Constructor.
         
         :param range: The range (+ or -) that the accelerometer will measure.
         :type range: :class:`.ADXL362.Range`
         :param port: The SPI port that the accelerometer is connected to
         :type port: :class:`.SPI.Port`
         """
+        super().__init__()
```
virtuald: Yeah.. since you keep adding them, let's pull those out and make them a separate PR. I agree the change should be made, it just needs to be added separately I think.

wpilib/wpilib/compressor.py
```diff
@@ -149,13 +149,7 @@ def clearAllPCMStickyFaults(self):
         """
         hal.clearAllPCMStickyFaults(self.module)
 
-    def initSendable(self, builder):
+    def initSendable(self, builder) -> None:
         builder.setSmartDashboardType("Compressor")
-        builder.addBooleanProperty("Enabled", self.enabled, self.enabledChanged)
+        builder.addBooleanProperty("Enabled", self.enabled, lambda value: self.start() if value else self.stop())
         builder.addBooleanProperty("Pressure switch", self.getPressureSwitchValue, None)
```
virtuald: Also not type hinting.

wpilib/wpilib/digitalglitchfilter.py
```diff
@@ -55,13 +58,13 @@ def _setFilter(input, channelIndex):
             # analog triggers are not supported for DigitalGlitchFilter
             if input.isAnalogTrigger():
                 raise ValueError("Analog Triggers are not supported for DigitalGlitchFilters")
-            hal.setFilterSelect(input.getPortHandleForRouting(), channelIndex)
```
virtuald: Again.

wpilib/wpilib/i2c.py
```diff
@@ -128,7 +128,7 @@ def write(self, registerAddress, data):
             return True
         return False
 
-    def writeBulk(self, data):
+    def writeBulk(self, data: [bytes]):
```
virtuald: Is that right?

wpilib/wpilib/i2c.py
```diff
@@ -182,7 +182,7 @@ def readOnly(self, count):
             raise ValueError("count must be at least 1, %s given" % count)
         return hal.readI2C(self.port, self.deviceAddress, count)
 
-    def verifySensor(self, registerAddress, expected):
+    def verifySensor(self, registerAddress: int, expected: [bytes]):
```
virtuald: Ditto.

wpilib/wpilib/i2c.py
auscompgeek: Yeah, this is definitely wrong.

wpilib/wpilib/adxl345_i2c.py
```diff
@@ -38,12 +39,12 @@ class ADXL345_I2C(SensorBase):
...
     Range = Accelerometer.Range
 
-    class Axes:
+    class Axes(IntEnum):
         kX = 0x00
         kY = 0x02
         kZ = 0x04
 
-    def __init__(self, port, range, address=None):
+    def __init__(self, port: I2C.port, range: Range, address: int=None):
```
auscompgeek: This is wrong. Please actually run the codebase against mypy. Also PEP 8.

wpilib/wpilib/doublesolenoid.py
```diff
@@ -5,13 +5,13 @@
 # must be accompanied by the FIRST BSD license file in the root directory of
 # the project.
 #----------------------------------------------------------------------------
+from .sendablebuilder import SendableBuilder
 
 import hal
-import enum
+from enum import IntEnum
```
auscompgeek: I'm not a fan of this change. This just means more lines we need to change when we inevitably decide to go and change everything to plain enums.

wpilib/wpilib/drive/differentialdrive.py
```diff
@@ -27,7 +29,7 @@ class DifferentialDrive(RobotDriveBase):
 
     Four motor drivetrain::
 
-        def robotInit(self):
+        def robotInit(self) -> None:
```
auscompgeek: I don't think this is really necessary. This should be kept as a simple example for teams.

auscompgeek:
Is type hinting the `initSendable` methods really necessary? They're not really intended to be directly called by teams, and actually type hinting them requires adding all the SendableBuilder imports.

wpilib/wpilib/i2c.py
twinters007: Oh yeah this is what I was asking about in gitter. It should just be `bytes`?

wpilib/wpilib/drive/differentialdrive.py
twinters007: Hah, my find/replace did this. Good catch

auscompgeek: That's horrifying. You could've used something like pyannotate or something instead.

virtuald:
Given that we already import everything anyways, I don't see any harm in it.

wpilib/wpilib/doublesolenoid.py
twinters007: yeah and I realize it's more pythonic to import enum and then show that you're getting IntEnum from enum anyway. I'm changing back

wpilib/wpilib/drive/differentialdrive.py
twinters007: Well with the amount of inconsistencies and fun python stuff we do, this was easier. And it worked great for the most part

wpilib/wpilib/doublesolenoid.py
twinters007: Although I'm going to leave this for a different PR anyway

twinters007:
If someone could help me with these circular imports that'd be much appreciated

auscompgeek:
Either take a read of PEP 484, or just don't have circular imports in the first place.

wpilib/wpilib/counter.py
```diff
@@ -413,14 +410,14 @@ def getDistance(self):
         """
```
auscompgeek: Please remove the extraneous return type documentation.

wpilib/wpilib/command/subsystem.py
```diff
@@ -43,7 +45,7 @@ def __init__(self, name=None):
             name = self.__class__.__name__
         self.setName(name)
 
-        Scheduler.getInstance().registerSubsystem(self)
+        scheduler.Scheduler.getInstance().registerSubsystem(self)
```
auscompgeek: please stop

wpilib/wpilib/digitalinput.py
```diff
@@ -57,37 +58,37 @@ def get(self):
         """
         return hal.getDIO(self.handle)
 
-    def getChannel(self):
+    def getChannel(self) -> int:
         """Get the channel of the digital input.
 
         :returns: The GPIO channel number that this object represents.
         :rtype: int
```
auscompgeek: Please remove the extraneous return type documentation.

wpilib/wpilib/digitalinput.py
```diff
@@ -57,37 +58,37 @@ def get(self):
...
         """
         return self.channel
 
-    def getAnalogTriggerTypeForRouting(self):
+    def getAnalogTriggerTypeForRouting(self) -> int:
         """Get the analog trigger type.
 
         :returns: false
         :rtype: int
```
auscompgeek: Ditto.

wpilib/wpilib/digitalinput.py
```diff
@@ -57,37 +58,37 @@ def get(self):
...
         """
         return 0
 
-    def isAnalogTrigger(self):
+    def isAnalogTrigger(self) -> bool:
         """Is this an analog trigger.
 
         :returns: true if this is an analog trigger
         :rtype: bool
```
auscompgeek: Ditto.

wpilib/wpilib/analogpotentiometer.py
```diff
@@ -5,6 +5,9 @@
 # must be accompanied by the FIRST BSD license file in the root directory of */
 # the project.                                                               */
 #----------------------------------------------------------------------------*/
+from typing import Union
+
+from .sendablebuilder import SendableBuilder
```
auscompgeek: Why is this import in a lone group?

wpilib/wpilib/analogtrigger.py
```diff
@@ -7,6 +7,9 @@
 # ----------------------------------------------------------------------------
 
 import weakref
+from typing import Union
+
+from .sendablebuilder import SendableBuilder
```
auscompgeek: Ditto.

wpilib/wpilib/builtinaccelerometer.py
```diff
@@ -4,6 +4,7 @@
 # Open Source Software - may be modified and shared by FRC teams. The code
 # must be accompanied by the FIRST BSD license file in $(WIND_BASE)/WPILib.
 #----------------------------------------------------------------------------
+from .sendablebuilder import SendableBuilder
```
auscompgeek: Ditto.

wpilib/wpilib/analogtriggeroutput.py
```diff
@@ -64,6 +64,7 @@ def __init__(self, trigger, outputType):
         :param outputType: An enum that specifies the output on the trigger
             to represent.
         """
+        super().__init__()
```
auscompgeek: This has nothing to do with type hinting.

wpilib/wpilib/compressor.py
```diff
@@ -55,31 +55,31 @@ def stop(self):
         """
         self.setClosedLoopControl(False)
 
-    def enabled(self):
+    def enabled(self) -> bool:
         """Get the enabled status of the compressor.
         
         :returns: True if the compressor is on
         :rtype: bool
```
auscompgeek: Please remove the extraneous return type documentation.

wpilib/wpilib/compressor.py
```diff
@@ -55,31 +55,31 @@ def stop(self):
...
         """
         return hal.getCompressor(self.compressorHandle)
 
-    def getPressureSwitchValue(self):
+    def getPressureSwitchValue(self) -> bool:
         """ Get the pressure switch value.
  
         :returns: True if the pressure is low
         :rtype: bool
```
auscompgeek: Ditto.

wpilib/wpilib/compressor.py
```diff
@@ -55,31 +55,31 @@ def stop(self):
...
         """
         return hal.getCompressorPressureSwitch(self.compressorHandle)
 
-    def getCompressorCurrent(self):
+    def getCompressorCurrent(self) -> float:
         """Get the current being used by the compressor.
         
         :returns: Current consumed by the compressor in amps
         :rtype: float
```
auscompgeek: Ditto.

wpilib/wpilib/compressor.py
```diff
@@ -88,57 +88,57 @@ def setClosedLoopControl(self, on):
         """
         hal.setCompressorClosedLoopControl(self.compressorHandle, True if on else False)
 
-    def getClosedLoopControl(self):
+    def getClosedLoopControl(self) -> bool:
         """Gets the current operating mode of the PCM.
         
         :returns: True if compressor is operating on closed-loop mode
         :rtype: bool
```
auscompgeek: Ditto.

wpilib/wpilib/compressor.py
```diff
@@ -88,57 +88,57 @@ def setClosedLoopControl(self, on):
         """
```
auscompgeek: You missed one.

wpilib/wpilib/controllerpower.py
```diff
@@ -59,7 +59,7 @@ def getCurrent3V3():
         return hal.getUserCurrent3V3()
 
     @staticmethod
-    def getEnabled3V3():
+    def getEnabled3V3() -> float:
```
auscompgeek: Huh? Is it a float, or is it a bool?

wpilib/wpilib/controllerpower.py
```diff
@@ -29,7 +29,7 @@ def getInputVoltage():
         return hal.getVinVoltage()
```
auscompgeek: Please remove the extraneous return type documentation. (I seem to keep saying this, and there's a lot in here so… just imagine a million dittos.)

wpilib/wpilib/counter.py
```diff
@@ -5,6 +5,7 @@
 # must be accompanied by the FIRST BSD license file in the root directory of
 # the project.
 #----------------------------------------------------------------------------
+from .sendablebuilder import SendableBuilder
```
auscompgeek: Why is this in a lone group of imports?

wpilib/wpilib/counter.py
```diff
@@ -465,15 +462,15 @@ def getStopped(self):
         """
         return hal.getCounterStopped(self.counter)
 
-    def getDirection(self):
+    def getDirection(self) -> bool:
         """The last direction the counter value changed.
 
         :returns: The last direction the counter value changed.
         :rtype: bool
```
auscompgeek: I think you're probably getting sick of this by now.

wpilib/wpilib/digitalglitchfilter.py
```diff
@@ -4,6 +4,9 @@
 # Open Source Software - may be modified and shared by FRC teams. The code
 # must be accompanied by the FIRST BSD license file in $(WIND_BASE)/WPILib.
 #----------------------------------------------------------------------------
+from typing import Union
+
+from .sendablebuilder import SendableBuilder
```
auscompgeek: Why is this in a lone group of imports?

wpilib/wpilib/digitaloutput.py
```diff
@@ -70,36 +72,36 @@ def free(self):
         hal.freeDIOPort(self.handle)
         self.handle = 0
 
-    def set(self, value):
+    def set(self, value: bool):
         """Set the value of a digital output.
 
         :param value: True is on, off is False
         :type  value: bool
```
auscompgeek: 20 review comments? I think it's time to bail now.

wpilib/wpilib/analogtriggeroutput.py
twinters007: Huh I thought I got them all

wpilib/wpilib/command/command.py
```diff
@@ -6,13 +6,21 @@
 # the project.
 #----------------------------------------------------------------------------
 
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
```
twinters007: I'm not proud of this, but seems to be the only way to get type hints

wpilib/wpilib/command/command.py
```diff
@@ -248,19 +256,19 @@ def isTimedOut(self):
...
 
-    def lockChanges(self):
+    def lockChanges(self) -> None:
         """Prevents further changes from being made
         """
         with self.mutex:
             self.locked = True
 
-    def setParent(self, parent):
+    def setParent(self, parent: 'commandgroup.CommandGroup') -> None:
```
auscompgeek: This type hint is absurd.

wpilib/wpilib/command/commandgroup.py
```diff
@@ -6,11 +6,11 @@
 # the project.
 #----------------------------------------------------------------------------
 
-from .command import Command
+from . import command
 
 __all__ = ["CommandGroup"]
 
-class CommandGroup(Command):
+class CommandGroup(command.Command):
```
auscompgeek: Why?

wpilib/wpilib/counter.py
```diff
@@ -207,16 +207,12 @@ def setUpSource(self, *args, **kwargs):
...
-        :type channel: int
         :param source: The digital source to count
-        :type source: DigitalInput
         :param analogTrigger:
             The analog trigger object that is used for the Up Source
-        :type analogTrigger: AnalogTrigger
         :param triggerType:
             The analog trigger output that will trigger the counter.
             Defaults to kState if not specified.
-        :type triggerType: AnalogTrigger.AnalogTriggerType
```
auscompgeek: Huh? Why did you delete these?

wpilib/wpilib/digitalsource.py
```diff
@@ -16,7 +16,7 @@
 __all__ = ["DigitalSource"]
 
 
-def _freeDigitalSource(handle):
+def _freeDigitalSource(handle: hal.DigitalHandle) -> None:
```
auscompgeek: Given that this is an internal function, I'm not too sure this was necessary.

wpilib/wpilib/counter.py
twinters007: Same reason I deleted the return types?

auscompgeek: Then where are the corresponding type hints?

wpilib/wpilib/command/commandgroup.py
twinters007: Circular imports. Command-based has a lot of circular dependency which doesn't work well with python. To include the type hints and keep the program working, I had to use a lot of module importing

wpilib/wpilib/digitalsource.py
twinters007: Makes mypy a little happier. And also makes reading the code easier

wpilib/wpilib/counter.py
twinters007: hehe classic find/replace at it again. I forgot about the vararg functions...\*sigh\*

wpilib/wpilib/command/commandgroup.py
auscompgeek: There's no circular import here at runtime.

wpilib/wpilib/counter.py
auscompgeek: I think you need to utilise find and replace better…

wpilib/wpilib/cameraserver.py
```diff
@@ -1,4 +1,6 @@
 # notrack
+from subprocess import Popen
```
auscompgeek: What? Just move the subprocess import up here if you're going to do this.

wpilib/wpilib/cameraserver.py
```diff
@@ -24,12 +26,12 @@ class CameraServer:
...
     
     @classmethod
-    def is_alive(cls):
+    def is_alive(cls) -> bool:
         ''':returns: True if the CameraServer is still alive'''
         return cls._alive
     
     @classmethod
-    def launch(cls, vision_py=None):
+    def launch(cls, vision_py: str = None) -> Optional[bool]:
```
auscompgeek: Where does this return a bool?

wpilib/wpilib/lineardigitalfilter.py
```diff
@@ -76,15 +80,12 @@ class LinearDigitalFilter(Filter):
     
     """
     
-    def __init__(self, source, ffGains, fbGains):
+    def __init__(self, source: PIDSource, ffGains: List[float], fbGains: List[float]) -> None:
```
auscompgeek: These type hints do not match the previously documented types.

wpilib/wpilib/command/command.py
twinters007: That's how you have to do it with `TYPE_CHECKING`

wpilib/wpilib/lineardigitalfilter.py
```diff
@@ -76,15 +80,12 @@ class LinearDigitalFilter(Filter):
     
     """
     
-    def __init__(self, source, ffGains, fbGains):
+    def __init__(self, source: PIDSource, ffGains: Union[List[float], Tuple[float]], fbGains: Union[List[float], Tuple[float]]) -> None:
```
auscompgeek: The type hint for `source` still doesn't match the original documented type.

Also, there's a much saner way of specifying the types for `ffGains` and `fbGains` which would make much more sense in this scenario.

twinters007: I bet it has to do with 'Iterable'

auscompgeek: You want a `Sequence` here, because `Iterable` won't type check.

wpilib/travis-requirements.txt
```diff
@@ -6,6 +6,6 @@ git+git://github.com/robotpy/pynetworktables.git
 
 git+git://github.com/robotpy/pyfrc.git
 
-pytest>=2.8
+pytest>=3.8
```
twinters007: 3.8 works with type hints, whereas 3.4 does not. (Not sure about .5-.7)

wpilib/wpilib/lineardigitalfilter.py
```diff
@@ -80,7 +80,7 @@ class LinearDigitalFilter(Filter):
     
     """
     
-    def __init__(self, source: PIDSource, ffGains: Union[List[float], Tuple[float]], fbGains: Union[List[float], Tuple[float]]) -> None:
+    def __init__(self, source: Union[PIDSource, Callable], ffGains: Sequence, fbGains: Sequence) -> None:
```
auscompgeek: A sequence of what?

auscompgeek:
Given there's so many merge conflicts now, we should probably close this and do this one step at a time. Convert the existing type information first (starting with the simpler ones that are automatable), then go through and add in the missing type information.

twinters007:
If I update merge conflicts and drop command stuff, it should pass tests and everything. Then we can work on command bit by bit. Probably not that important for robotpy anyway

twinters007:
But the changes will have to come after December 15th when school gets out

wpilib/wpilib/ultrasonic.py
```diff
@@ -5,16 +5,21 @@
...
+from enum import IntEnum
+from typing import Union
 
 import hal
 import threading
 import weakref
 
 from .counter import Counter
 from .interfaces import PIDSource
+from .digitalinput import DigitalInput
```
twinters007: Looks to be an artifact of black

wpilib/travis-requirements.txt
```diff
@@ -6,6 +6,6 @@ git+git://github.com/robotpy/pynetworktables.git
 
 git+git://github.com/robotpy/pyfrc.git
 
-pytest>=2.8
+pytest<4.0.0
```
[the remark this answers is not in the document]
virtuald: Well, we should fix that then. :)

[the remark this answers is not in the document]
virtuald: There was an issue with pyfrc doing that... I think I addressed it?

wpilib/wpilib/analogaccelerometer.py
```diff
@@ -5,17 +5,19 @@
...
 from .interfaces import PIDSource
-from .sendablebase import SendableBase
+from .sensorbase import SensorBase
+from .sendablebuilder import SendableBuilder
 
 __all__ = ["AnalogAccelerometer"]
 
 
-class AnalogAccelerometer(SendableBase):
+class AnalogAccelerometer(SensorBase):
```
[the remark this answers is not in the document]
twinters007: 🤷‍♀️ 
fixing

wpilib/wpilib/analogtrigger.py
```diff
@@ -38,7 +40,7 @@ class AnalogTrigger(SendableBase):
 
     AnalogTriggerType = AnalogTriggerOutput.AnalogTriggerType
```
auscompgeek: You should be able to fix the cyclic import by moving this here.

wpilib/wpilib/pwm.py
```diff
@@ -119,37 +120,38 @@ def stopMotor(self):
...
-        :type center: float
         :param deadbandMin: The low end of the deadband pulse width in ms
-        :type deadbandMin: float
         :param min: The minimum pulse width in ms
-        :type min: float
         """
         hal.setPWMConfig(self.handle, max, deadbandMax, center, deadbandMin, min)
 
-    def getRawBounds(self):
+    def getRawBounds(self) -> Tuple[(int, int, int, int, int)]:
```
auscompgeek: nit: unnecessary parens around tuple

wpilib/wpilib/safepwm.py
```diff
@@ -0,0 +1,42 @@
+# validated: 2017-10-07 EN e1195e8b9dab edu/wpi/first/wpilibj/SafePWM.java
```
auscompgeek: Why did you create this file?

wpilib/wpilib/pidcontroller.py
```diff
@@ -447,7 +442,7 @@ def setToleranceBuffer(self, bufLength):
             self.filter = LinearDigitalFilter.movingAverage(self.origSource, bufLength)
             self.pidInput = self.filter
 
-    def onTarget(self):
+    def onTarget(self) -> Optional[bool]:
```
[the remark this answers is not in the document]
twinters007: Whoops, didn’t mean to resolve

It’s a mypy thing. onTarget doesn’t return a Boolean until a tolerance value is set. If we don’t care about mypy, we can leave it as bool

wpilib/wpilib/analogtrigger.py
twinters007: Moving what?

wpilib/wpilib/safepwm.py
twinters007: I thought I got rid of that....hmm

It came from when I merged

wpilib/wpilib/pidcontroller.py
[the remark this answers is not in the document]
auscompgeek: If we're really gonna keep walking this path, maybe we should just assign this dummy in `__init__` instead. That should keep mypy happy.

wpilib/wpilib/adxrs450_gyro.py
```diff
@@ -43,12 +43,11 @@ class ADXRS450_Gyro(GyroBase):
     kSNHighRegister = 0x0E
     kSNLowRegister = 0x10
 
-    def __init__(self, port=None):
+    def __init__(self, port: SPI.Port = None) -> None:
```
virtuald: Shouldn't this be `Optional[SPI.Port]`?

wpilib/wpilib/adxl362.py
```diff
@@ -49,13 +51,11 @@ class Axes:
         kY = 0x02
         kZ = 0x04
 
-    def __init__(self, range, port=None):
+    def __init__(self, range: Range, port: SPI.Port = None) -> None:
```
virtuald: Optional

wpilib/wpilib/analoggyro.py
```diff
@@ -37,7 +38,9 @@ class AnalogGyro(GyroBase):
 
     PIDSourceType = PIDSource.PIDSourceType
 
-    def __init__(self, channel, center=None, offset=None):
+    def __init__(
+        self, channel: Union[AnalogInput, int], center: int = None, offset: float = None
```
virtuald: Two optionals.

wpilib/wpilib/command/pidsubsystem.py
```diff
@@ -23,7 +23,15 @@ class PIDSubsystem(Subsystem):
...
 
-    def __init__(self, p, i, d, period=None, f=0.0, name=None):
+    def __init__(
+        self,
+        p: float,
+        i: float,
+        d: float,
+        period: float = None,
+        f: float = 0.0,
+        name: str = None,
```
virtuald: Optional.

wpilib/wpilib/command/subsystem.py
```diff
@@ -32,7 +37,7 @@ class Subsystem(SendableBase):
     .. seealso:: :class:`.Command`
     """
 
-    def __init__(self, name=None):
+    def __init__(self, name: str = None) -> None:
```
virtuald: Optional

wpilib/wpilib/command/subsystem.py
```diff
@@ -145,7 +150,7 @@ def getCurrentCommandName(self):
             return currentCommand.getName()
         return ""
 
-    def addChild(self, child, name=None):
+    def addChild(self, child: "sendable.Sendable", name: str = None) -> None:
```
virtuald: Optional.

wpilib/wpilib/i2c.py
```diff
@@ -35,11 +35,10 @@ class Port:
         kOnboard = 0
         kMXP = 1
 
-    def __init__(self, port, deviceAddress, simPort=None):
+    def __init__(self, port: Port, deviceAddress: int, simPort: object = None) -> None:
```
virtuald: Optional.

wpilib/wpilib/interfaces/gamepadbase.py
```diff
@@ -20,48 +20,48 @@ class GamepadBase(GenericHID):
...
+    def getBumper(self, hand: GenericHID.Hand) -> bool:
         """Is the bumper pressed.
 
         :param hand: which hand
         :returns: true if the bumper is pressed
         """
         raise NotImplementedError
 
-    def getStickButton(self, hand=None):
+    def getStickButton(self, hand: GenericHID.Hand = None) -> bool:
```
virtuald: Optional.

wpilib/wpilib/interfaces/joystickbase.py
```diff
@@ -20,15 +20,15 @@ class JoystickBase(GenericHID):
...
+    def __init__(self, port: int) -> None:
         warnings.warn(
             "JoystickBase is deprecated. Inherit directly from GenericHID instead",
             DeprecationWarning,
             stacklevel=2,
         )
         super().__init__(port)
 
-    def getZ(self, hand=None):
+    def getZ(self, hand: GenericHID.Hand = None) -> float:
```
virtuald: Optional.

wpilib/wpilib/resource.py
```diff
@@ -74,7 +72,7 @@ def __init__(self, size):
         Resource._resource_objects.append(self)
         self.numAllocated = [None] * size
 
-    def allocate(self, obj, index=None):
+    def allocate(self, obj: object, index: int = None) -> int:
```
virtuald: Optional.

twinters007:
I think I got all the optionals

wpilib/wpilib/adxrs450_gyro.py
auscompgeek: It already is.

wpilib/wpilib/i2c.py
```diff
@@ -35,11 +36,12 @@ class Port:
         kOnboard = 0
         kMXP = 1
 
-    def __init__(self, port, deviceAddress, simPort=None):
+    def __init__(
+        self, port: Port, deviceAddress: int, simPort: Optional[object] = None
```
auscompgeek: I'm not sure `object` is an appropriate type hint here. hm...

twinters007: I mean, the best thing would be an interface, but it's not even a full implementation of an interface, just the methods you use. Do you have any recommendations?

auscompgeek: Honestly, I would just leave that argument untyped. I don't see a nice way to do this.

wpilib/wpilib/interruptablesensorbase.py
```diff
@@ -34,25 +34,25 @@ def __init__(self):
...
-    def getAnalogTriggerTypeForRouting(self):
+    def getAnalogTriggerTypeForRouting(self) -> int:
         raise NotImplementedError
 
-    def getPortHandleForRouting(self):
+    def getPortHandleForRouting(self) -> int:
         raise NotImplementedError
 
-    def requestInterrupts(self, handler=None):
+    def requestInterrupts(self, handler: Optional[Callable] = None) -> None:
```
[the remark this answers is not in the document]
twinters007: oh that's what it is...lemme see if adding types to callable fixes it

wpilib/wpilib/spi.py
```diff
@@ -193,7 +193,7 @@ def setChipSelectActiveLow(self) -> None:
         """Configure the chip select line to be active low."""
         hal.setSPIChipSelectActiveLow(self.port)
 
-    def write(self, dataToSend: bytes) -> int:
+    def write(self, dataToSend: Union[bytes, List[int]]) -> int:
```
auscompgeek: If we're going to do this, shouldn't this be `Iterable[int]`? Or do we not want to guarantee that? If not, `Sequence[int]` would be more appropriate.

wpilib/wpilib/spi.py
```diff
@@ -236,7 +236,7 @@ def read(self, initiate: bool, size: int) -> bytes:
         else:
             return hal.readSPI(self.port, size)
 
-    def transaction(self, dataToSend: bytes) -> bytes:
+    def transaction(self, dataToSend: bytes) -> Union[bytes, List[int]]:
```
auscompgeek: uhhhh we should probably change the HAL to return bytes here

ArchdukeTim: I totally changed the wrong thing. I meant to change the parameter, not return value

wpilib/wpilib/interruptablesensorbase.py
```diff
@@ -5,7 +5,7 @@
 # must be accompanied by the FIRST BSD license file in the root directory of
 # the project.
 # ----------------------------------------------------------------------------
-
+from typing import Callable, Optional, TypeVar
```
auscompgeek: `typing.TypeVar: unused import`

wpilib/wpilib/livewindow.py
```diff
@@ -156,8 +163,8 @@ def addSensor(cls, subsystem, name, component):
             component.setName(subsystem, name)
 
     @classmethod
-    def addActuator(cls, subsystem, name, component):
-        """Add an Actuator associated with the subsystem and with call it by
+    def addActuator(cls, subsystem, name: str, component: Sendable) -> None:
+        """Add an Actuator associated with the subsystsubsystem.Subsystemem and with call it by
```
auscompgeek: what

wpilib/wpilib/preferences.py
```diff
@@ -68,19 +69,19 @@ def __init__(self):
 
         hal.report(hal.UsageReporting.kResourceType_Preferences, 0)
 
-    def getKeys(self):
+    def getKeys(self) -> [str]:
```
auscompgeek: ahem

wpilib/wpilib/smartdashboard.py
```diff
@@ -380,12 +378,11 @@ def setDefaultBooleanArray(cls, key, defaultValue):
         return table.setDefaultBooleanArray(key, defaultValue)
 
     @classmethod
-    def getBooleanArray(cls, key, defaultValue):
+    def getBooleanArray(cls, key: str, defaultValue: List[bool]) -> List[bool]:
```
auscompgeek: NetworkTables will currently never return a list. Better to specify `Sequence[bool]`.

wpilib/wpilib/spi.py
```diff
@@ -317,8 +317,8 @@ def forceAutoRead(self) -> None:
         hal.forceSPIAutoRead(self.port)
 
     def readAutoReceivedData(
-        self, buffer, numToRead: int, timeout: float
-    ) -> (int, bytes):
+        self, buffer: bytes, numToRead: int, timeout: float
+    ) -> Tuple[(int, bytes)]:
```
auscompgeek: unnecessary parens

wpilib/wpilib/sendablebase.py
```diff
@@ -104,15 +104,15 @@ def setName(self, *args, **kwargs):
...
         with self.mutex:
             return self.subsystem
 
-    def setSubsystem(self, subsystem):
+    def setSubsystem(self, subsystem: str) -> None:
         with self.mutex:
             self.subsystem = subsystem
 
-    def addChild(self, child):
+    def addChild(self, child: object) -> None:
```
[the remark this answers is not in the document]
ArchdukeTim: upstream is `java.lang.Object`

wpilib/wpilib/smartdashboard.py
ArchdukeTim: I'm gonna use Iterable because networktables has type hints for iterable in some places

wpilib/wpilib/livewindow.py
```diff
@@ -164,7 +164,7 @@ def addSensor(
 
     @classmethod
     def addActuator(cls, subsystem, name: str, component: Sendable) -> None:
```
auscompgeek: shouldn't `subsystem` have a type hint here?

wpilib/wpilib/smartdashboard.py
```diff
@@ -378,7 +379,7 @@ def setDefaultBooleanArray(cls, key: str, defaultValue: List[bool]) -> bool:
         return table.setDefaultBooleanArray(key, defaultValue)
 
     @classmethod
-    def getBooleanArray(cls, key: str, defaultValue: List[bool]) -> List[bool]:
+    def getBooleanArray(cls, key: str, defaultValue: Iterable[bool]) -> Iterable[bool]:
```
auscompgeek: I don't think this is the contract we want to give to teams. This seems bad.

wpilib/wpilib/smartdashboard.py
auscompgeek: Last I checked, NetworkTables doesn't even have type hints?

wpilib/wpilib/sendablebase.py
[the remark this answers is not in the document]
auscompgeek: Huh, so it is.

wpilib/wpilib/smartdashboard.py
ArchdukeTim: defaultValue *is* of type iterable (ex. https://github.com/robotpy/pynetworktables/blob/master/networktables/networktable.py#L668), and it seems that we always return a Tuple, unless the key isn't found in which case we return the defaultValue (which is iterable)

auscompgeek: Bleh. I guess we should fix those. Would you agree that being able to index an array value is useful?

ArchdukeTim: mmmm yeah I do. I'll change it to sequence here since it isn't that important anyway

auscompgeek: As an aside, I think specifying _any_ arbitrary default value should be legal. Imagine a team that wants to handle missing values specially, say, having a default that is expensive to compute, so they pass in None and check whether the get returns None.

wpilib/wpilib/smartdashboard.py
virtuald: I think it would be better to specify Tuple explicitly, because it will always return a Tuple as that's immutable.

virtuald: The input default value could be anything -- but the return value will always be a tuple, and I'm perfectly happy with making that API.

ArchdukeTim: the return won't be a Tuple if it returns the default value though

auscompgeek: typing.Tuple[] would be incorrect - a Tuple has a fixed length. A Sequence[T] is an arbitrary length sequence (i.e. implements `__len__` and `__getitem__`) of T, which is correct.

auscompgeek: If we want to denote that the default value can be any type, it would be appropriate to have a type variable T, which is the type of the default value, and return a Union[T, Sequence[...]].

ArchdukeTim:
You’re too kind 😂

wpilib/wpilib/smartdashboard.py
```diff
@@ -77,11 +83,8 @@ def putData(cls, *args, **kwargs):
         - value
         
         :param key: the key (cannot be None)
-        :type  key: str
         :param data: the value
-        :type data: :class:`.Sendable`
         :param value: the value
-        :type value: :class:`.Sendable`
```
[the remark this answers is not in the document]
ArchdukeTim: I put this back. I probably got rid of it in my mass find/replace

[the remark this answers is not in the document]
auscompgeek: They are still missing.

wpilib/wpilib/sendablechooser.py
```diff
@@ -44,15 +45,15 @@ class SendableChooser(SendableBase):
...
+    def __init__(self) -> None:
         """Instantiates a SendableChooser.
         """
         super().__init__(addLiveWindow=False)
         self.map = {}
         self.tableSelected = None
         self.defaultChoice = ""
 
-    def addObject(self, name, object):
+    def addObject(self, name: str, object: object) -> None:
```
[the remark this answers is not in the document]
auscompgeek: It would make sense to do here what upstream does.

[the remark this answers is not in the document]
ArchdukeTim: with generics?

wpilib/wpilib/smartdashboard.py
```diff
@@ -77,7 +85,7 @@ def putData(cls, *args, **kwargs):
         - value
         
         :param key: the key (cannot be None)
-        :type  key: str
+        :type key: str
```
auscompgeek: Not too sure why you changed this line, but ok.

wpilib/wpilib/smartdashboard.py
```diff
@@ -395,7 +397,7 @@ def getBooleanArray(cls, key, defaultValue):
         return table.getBooleanArray(key, defaultValue)
 
     @classmethod
-    def putNumberArray(cls, key, value):
+    def putNumberArray(cls, key: str, value: List[float]) -> bool:
```
auscompgeek: This should be a Sequence rather than a List.

wpilib/wpilib/sendablechooser.py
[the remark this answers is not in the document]
auscompgeek: You forgot to bound the generic.

wpilib/wpilib/smartdashboard.py
ArchdukeTim: I'm surprised black didn't take care of it anyway

auscompgeek: black doesn't format docstrings. There are too many different docstring conventions to make that possible for a one-size-fits-all tool.

virtuald:
Are we there yet?

wpilib/wpilib/pidcontroller.py
```diff
@@ -9,12 +9,15 @@
 from collections import deque
 from itertools import islice
 import threading
+from typing import Optional, NoReturn, Union
```
auscompgeek: You do use pyflakes, right? Right??

auscompgeek:
@virtuald Are you fine with making the type signatures in SmartDashboard API?

wpilib/wpilib/pidcontroller.py
ArchdukeTim: It's hard to just use Pyflakes because it'll mark things that I didn't put in, and then I'd have to sift through the results lest you and dustin asking why I removed something that had nothing to do with type hints

virtuald: Sounds like we need to automatically run pyflakes...

virtuald:
I think as long as we're not verifying the signatures automatically via mypy, that getting close enough is good enough. We can fix things piecemeal after this PR.

wpilib/wpilib/smartdashboard.py
```diff
@@ -380,12 +383,11 @@ def setDefaultBooleanArray(cls, key, defaultValue):
         return table.setDefaultBooleanArray(key, defaultValue)
 
     @classmethod
-    def getBooleanArray(cls, key, defaultValue):
+    def getBooleanArray(cls, key: str, defaultValue: T) -> Union[T, Tuple[bool]]:
```
auscompgeek: I just realised this return types are very, very wrong.

auscompgeek:
To me, "close enough" would be saying that we're only guaranteeing that putting a sequence on NetworkTables will work. Iterable seems pedantic and potentially confusing. Do we want to make the guarantee that iterables can always be put onto NetworkTables?

ArchdukeTim:
@auscompgeek when it puts things in networktables, it iterates over them and turns it into a tuple. The final product is a tuple, but any iterable will be able to be parsed

wpilib/wpilib/smartdashboard.py
ArchdukeTim: What's wrong with them?

auscompgeek: The return value isn't necessarily a 1-tuple containing a bool.

auscompgeek:
Yes, but a sequence is also iterable. We also haven't fully defined the types throughout all of pynetworktables (a bunch of which the documented types are wrong anyway).

My question still stands.

virtuald:
The implementation is at https://github.com/robotpy/pynetworktables/blob/59b5d1328274e78393202fe2a7e09845f5ff63a5/ntcore/value.py#L57 , using an iterable seems best?

wpilib/wpilib/smartdashboard.py
ArchdukeTim: How is it not?

ArchdukeTim: @virtuald do you have any insight?

auscompgeek: If I set a NT to an array of 5 booleans, I wouldn't expect to get back a tuple containing only one boolean. Would you?

virtuald: Oh, I see what you mean. Sigh. 

[Looks like `Tuple[bool, ...]` would work?](https://github.com/python/mypy/issues/184)

auscompgeek: If you're fine with guaranteeing that a tuple is returned, that works.

wpilib/wpilib/sendablechooser.py
```diff
@@ -5,8 +5,12 @@
 # must be accompanied by the FIRST BSD license file in the root directory of
 # the project.
 # ----------------------------------------------------------------------------
+from typing import TypeVar
+
+V = TypeVar("V")
```
auscompgeek: This type variable isn't particularly useful if it doesn't tie any types together.

If this is intended to tie together the argument and return types for SendableChooser's methods, you might also want to make this type variable covariant.

(i.e. if I have a SendableChooser of SpeedControllers, I should be able to put in a Spark, get it back, and treat it as merely a SpeedController.)

ArchdukeTim: Wouldn't a spark fall under SpeedController since it's a subclass? I add a SpeedController object (which happens to be a spark), and I get back a speedcontroller object (that is *also* a spark)

auscompgeek: We appear to be in violent agreement with each other.

ArchdukeTim: Oh I didn't realize python had a special marking for this type of behavior. All these small quirks have made me wonder if I should've read through pep484 before implementing it...🤔

wpilib/wpilib/sendablechooser.py
```diff
@@ -5,8 +5,12 @@
 # must be accompanied by the FIRST BSD license file in the root directory of
 # the project.
 # ----------------------------------------------------------------------------
+from typing import TypeVar
+
+V = TypeVar("V", covariant=True)
```
auscompgeek: FYI, you're still just using this as a convoluted way of saying Any.

ArchdukeTim: Yeah if I said `Any` that wouldn't relate the classes

ArchdukeTim: Oh I guess you can add multiple types to a chooser so ok.

auscompgeek: You had no relationship between them anyway?

ArchdukeTim: Yeah, before if I put in type V, I was gonna get a type V back

auscompgeek: V was a free variable. i.e. in each signature, V did not relate to any V in any other signature.

virtuald:
Any other nitpicks, make a new PR.

PR #472, opened 2018-05-25
Use enum for enum classes

twinters007:


wpilib/wpilib/iterativerobotbase.py
```diff
@@ -47,7 +47,7 @@ class IterativeRobotBase(RobotBase):
     - testPeriodic()
     """
 
-    class Mode(enum.Enum):
+    class Mode(enum.IntEnum):
```
auscompgeek: Wouldn't it be better to keep this as a plain Enum?

wpilib/wpilib/command/commandgroup.py
```diff
@@ -5,6 +5,7 @@
 # must be accompanied by the FIRST BSD license file in the root directory of
 # the project.
 #----------------------------------------------------------------------------
+import enum
```
auscompgeek: Unused import.

wpilib/wpilib/serialport.py
```diff
@@ -35,32 +36,32 @@ class SerialPort:
...
         kSpace = 4
             
-    class StopBits:
+    class StopBits(enum.IntEnum):
         kOne = 10
         kOnePointFive = 15
         kTwo = 20
             
-    class FlowControl:
+    class FlowControl(enum.IntFlag):
```
auscompgeek: In case you didn't realise already from the CI logs, `enum.Flag` is new in Python 3.6. Unless we drop support for Python 3.5, we can't use it.

twinters007: Sounds like....we should...drop support 👍

auscompgeek: You'll have to give some compelling reasons to drop support for Python 3.5 though.

- Debian stable (stretch) ships Python 3.5.
- Ubuntu 16.04 (the second-latest LTS release, which is still supported) ships with Python 3.5.

auscompgeek:
What's the status on this?

twinters007:
¯\\\_(ツ)_/¯ Probably should wait on more changes to come through. This is pretty minor, so I'll close this one and we can add them in when there are fewer conflicts

PR #478, opened 2018-09-10
update spi for 2019

ariovistus:


auscompgeek:
This has broken builds for anything that tests the navx package. Could someone please go and fix that?

PR #479, opened 2018-09-10
update counter for 2019

ariovistus:


virtuald:
I think this needs a rebase.

ariovistus:
it still won't pass until digitalsource or digitaloutput is updated

PR #481, opened 2018-09-10
update SendableBase for 2019

ariovistus:


wpilib/wpilib/sendablebase.py
```diff
@@ -34,6 +35,15 @@ def __init__(self, addLiveWindow=True):
...
     def free(self):
+        """
+        .. deprecated:: 2019.0.0
+            Use close instead
+        """
+        warnings.warn("use close instead",
+                      DeprecationWarning, stacklevel=2)
+        self.close()
+
+    def close(self):
```
virtuald: Heh, that's interesting. If that's the route to take, then our internal free stuff should probably use close instead.

auscompgeek: I personally think we should just ignore this commit. Unless if there's some weird IDE out there that understands the semantics of a `close` method, this seems like a bad idea. Even then, this still sounds like a bad idea.

ariovistus: does `free` have particular semantics in python? I assume `AutoCloseable` does in java. I thought about ignoring this, but they changed free everywhere in wpilib, and the source difference would make things confusing

virtuald: I'm inclined to copy what they do, and switch free/close. Probably do both for one year, and then get rid of free.

Why is this a bad idea?

auscompgeek: I can't entirely remember why I thought this was a bad idea, but it probably has something to do with the semantic meaning of `close`.

Come to think of it, we don't use finalisers where upstream does either.

virtuald:
Fix the tests (docstring issue it seems) and let's merge this. We'll switch to close/free, with the intention of removing free next year.

PR #482, opened 2018-09-10
Add python 3.7 to tests

virtuald:


auscompgeek:
I wonder if we should start using Circle CI instead.

auscompgeek:
Bleh. That didn't run 3.7. Let's try the other way of doing multiple jobs.

auscompgeek:
Is declaring a build matrix and build stages mutually exclusive? Hmm...

ariovistus:
not in my experience, e..g. https://github.com/ariovistus/pyd/blob/master/.travis.yml

virtuald:
Looks like it ran in 3.7 this time?

auscompgeek:
@ariovistus That doesn't look like it uses matrix.include.

PR #483, opened 2018-09-10
can haz docker ci?

ariovistus:
probably ignore this, travis is being dumb and I can't enable it on my fork

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,19 @@
 language: python
-sudo: false
+sudo: required
+services: docker
 stages:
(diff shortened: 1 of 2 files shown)


ariovistus:
oh hey it does work

virtuald:
This seems worse. Why not just enable docker for the one that requires it (3.7), that way once they fix it we can go back to their setup?

... also, I like docker, but weren't there other solutions posted somewhere?

ariovistus:
fair enough. I got my lols.

PR #490, opened 2018-09-19
update notifier for 2019

ariovistus:


ariovistus:
probably depends on resource update

wpilib/wpilib/notifier.py
```diff
@@ -38,15 +38,17 @@ def __init__(self, run: callable) -> None:
...
         self._thread.start()
         
         # python-specific
         Resource._add_global_resource(self)
     
-    # python-specific: this is called finalize in Java
-    def free(self) -> None:
+    def close(self) -> None:
         handle, self._notifier = self._notifier, None
+        if handle == 0:
```
virtuald: This should be `if not handle`?

wpilib/wpilib/resource.py
```diff
@@ -40,15 +40,15 @@ def _reset():
                 if ref is None:
                     continue
                 obj = ref()
-                if obj is not None and hasattr(obj, 'free'):
-                    obj.free()
+                if obj is not None and hasattr(obj, 'close'):
```
virtuald: This needs to allow free and close.

ariovistus: if both are defined, can close handle being called twice?

virtuald: It should. But, I think just call one or the other, preferring close.

wpilib/wpilib/resource.py
```diff
@@ -40,14 +40,18 @@ def _reset():
...
                     obj.free()
             
             resource.numAllocated = [None]*len(resource.numAllocated)
         
         for ref in Resource._global_resources:
             obj = ref()
-            if obj is not None and hasattr(obj, 'free'):
+            if obj is not None and hasattr(obj, 'close'):
+                obj.close()
+            elif obj is not None and hasattr(obj, 'free'):
```
auscompgeek: Maybe refactor these to not repeat `obj is not None` as well.

auscompgeek:
Needs updating again and verification.

ariovistus:
upstream added a setUncaughtExceptionHandler here: https://github.com/wpilibsuite/allwpilib/blob/master/wpilibj/src/main/java/edu/wpi/first/wpilibj/Notifier.java#L124

I'm assuming we're going to do what dustin wants: https://github.com/wpilibsuite/allwpilib/pull/926
?

auscompgeek:
Yeah, I don't see catching uncaught exceptions being particularly useful in Notifier for us. Threads should exit cleanly and die if the main thread dies for us anyway.

wpilib/wpilib/notifier.py
```diff
@@ -77,6 +92,8 @@ def _run(self) -> None:
                 if self._periodic:
                     self._expirationTime += self._period
                     self._updateAlarm()
+                else:
+                    self._updateAlarm(-1)
```
auscompgeek: You should add in the comment from upstream here explaining what this is for.

ariovistus: I read that. And didn't comprehend it. And still don't. I'm kinda tired..

PR #491, opened 2018-09-19
update digitalglitchfilter for 2019

ariovistus:


ariovistus:
```
type object 'UsageReporting' has no attribute 'kResourceType_DigitalGlitchFilter'
```
waiting on hal I think

auscompgeek:
You should port over that change then :)

PR #502, opened 2018-09-19
update analog and builtin accelerometers for 2019

ariovistus:


wpilib/wpilib/builtinaccelerometer.py
```diff
@@ -88,3 +88,6 @@ def initSendable(self, builder):
         builder.addDoubleProperty("X", self.getX, None)
         builder.addDoubleProperty("Y", self.getY, None)
         builder.addDoubleProperty("Z", self.getZ, None)
+
+    def getSimObject(self):
```
virtuald: Hm?

ariovistus: I don't know what it is

auscompgeek: We should probably add that commit (`337e89cf6e791d67f6e80fbba8c613cc98d2a36a`) to the git source-track ignore file... thoughts @virtuald?

PR #505, opened 2018-09-30
update pidcontroller for 2019

ariovistus:
depends on #490

Fixes: #38

### DIFF
--- a/wpilib/wpilib/pidbase.py
+++ b/wpilib/wpilib/pidbase.py
@@ -1,0 +1,501 @@
+# validated: 2019-01-02 EN 89d15f061b33 edu/wpi/first/wpilibj/PIDBase.java
+# ----------------------------------------------------------------------------
+# Copyright (c) FIRST 2008-2018. All Rights Reserved.
+# Open Source Software - may be modified and shared by FRC teams. The code
+# must be accompanied by the FIRST BSD license file in the root directory of
+# the project.
+# ----------------------------------------------------------------------------
+
+import threading
+import warnings
+
+import hal
+
+from .sendablebuilder import SendableBuilder
+from .interfaces.pidsource import PIDSource
+from .sendablebase import SendableBase
+from .resource import Resource
+from .timer import Timer
+from .lineardigitalfilter import LinearDigitalFilter
+
+__all__ = ["PIDBase"]
+
+
+class PIDBase(SendableBase):
+    """Class implements a PID Control Loop.
+
+    Creates a separate thread which reads the given PIDSource and takes care of the integral
+    calculations, as well as writing the given PIDOutput.
+
+    This feedback controller runs in discrete time, so time deltas are not used in the integral
+    and derivative calculations. Therefore, the sample rate affects the controller's behavior for a
+    given set of PID constants.
+
+    """
+
+    kDefaultPeriod = 0.05
+    instances = 0
+
+    PIDSourceType = PIDSource.PIDSourceType
+
+    # Tolerance is the type of tolerance used to specify if the PID controller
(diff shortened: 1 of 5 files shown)


wpilib/wpilib/pidbase.py
```diff
@@ -0,0 +1,501 @@
...
+from .resource import Resource
+from .timer import Timer
+from .lineardigitalfilter import LinearDigitalFilter
+
+__all__ = ["PIDBase"]
+
+class PIDBase(SendableBase):
+    """Class implements a PID Control Loop.
+
+    Creates a separate thread which reads the given PIDSource and takes care of the integral
```
auscompgeek: Heh, this seems wrong; this class doesn't spawn a thread. But it's what upstream has. I guess we should file a bug, I suppose.

ariovistus: my impression was it leaves the choice of notifier vs non notifier up to the subclass

auscompgeek: I think that's the intention here, yes. But that doesn't change the fact that there's a documentation bug here.

auscompgeek:
Rebase this and we'll merge, I guess.

wpilib/wpilib/pidbase.py
```diff
@@ -0,0 +1,500 @@
...
+        """
+        Wraps error around for continuous inputs. The original error is
+        returned if continuous mode is disabled. This is an unsynchronized
+        function.
+
+        :param error: The current error of the PID controller.
+        :return: Error for continuous inputs.
+        """
+        if self.continuous and self.inputRange > 0:
+            error %= self.inputRange
```
auscompgeek: You know, I keep staring at this line. I know you didn't originally write this, but the meaning of `%` in Python is different to other C-like languages, so `error` is always positive here. The rough equivalent of C/Java's `%` is `math.fmod()`.

I sat down at 1am thinking this out, and I'm pretty sure the following is equivalent to both what upstream is doing and what this is doing:

```python
error %= self.inputRange
if error > self.inputRange / 2:
    return error - self.inputRange
```

ariovistus: Pretty sure I looked at that and wondered about it, but figured "meh, it's in the existing codebase, it's probably fine"

auscompgeek: Yeah. I'm surprised we haven't ported across the PID tolerance tests.

wpilib/wpilib/pidbase.py
```diff
@@ -0,0 +1,500 @@
...
+    # is on target.  The various implementations of this such as
+    # PercentageTolerance and AbsoluteTolerance specify types of tolerance
+    # specifications to use.
+    def PercentageTolerance_onTarget(self, percentage):
+        return abs(self.getError()) < percentage / 100.0 * self.inputRange
+
+    def AbsoluteTolerance_onTarget(self, value):
+        return abs(self.getError()) < value
+
+    def __init__(self, Kp, Ki, Kd, Kf, output, source):
```
auscompgeek: Java has a default for Kf here. Do you think it's worth putting it here too? We can make source and output keyword-only arguments.

Wait, source and output are in the wrong order compared to Java too.

ariovistus: I think it's not worth putting on the base

auscompgeek: The argument order still needs to be fixed though.

Having the default here could simplify the PIDController constructor too.

auscompgeek:
Needs another rebase since we merged the type hints PR. Please ensure the type hints are preserved in PIDBase as well.

PR #508, opened 2018-09-30
update dmc60 for 2019

ariovistus:


ariovistus:
looks like its waiting on an update to hal

PR #509, opened 2018-09-30
update driver station for 2019

ariovistus:


wpilib/tests/test_driverstation.py
```diff
@@ -109,8 +109,6 @@ def test_getStickAxis_limits(dsmock, halmock):
         dsmock.getStickAxis(-1, 1)
     with pytest.raises(IndexError):
         dsmock.getStickAxis(dsmock.kJoystickPorts, 1)
-    with pytest.raises(IndexError):
-        dsmock.getStickAxis(0, -1)
```
ariovistus: not sure why upstream removed the check

wpilib/wpilib/driverstation.py
```diff
@@ -118,6 +118,7 @@ def __init__(self):
         if not hasattr(DriverStation, 'instance') or DriverStation.instance is not None:
             raise ValueError("Do not create DriverStation instances, use DriverStation.getInstance() instead")
 
+        hal.initialize(500, 0)
```
auscompgeek: Erm. If you're actually going to do this, you should also remove the corresponding call in the hal package itself. But I'm not sure we should do this.

wpilib/tests/test_driverstation.py
auscompgeek: Hm. We should ask whether that was intentional. I suspect not.

wpilib/wpilib/driverstation.py
virtuald: I would have to look closer, but I'm 98% sure we would not want to do this here. Probably best to comment it out and add a comment saying that we're intentionally leaving it out.

auscompgeek: Thad mentioned upstream added it here to help with unit testing, since Java doesn't really have an equivalent of ensuring some code gets executed on first import. Of course, we ensure the HAL is initialised whenever it's imported, so this shouldn't be a concern for us.

wpilib/tests/test_driverstation.py
virtuald: Upstream is incorrect, we should make ours correct.

virtuald:
Bug filed upstream at https://github.com/wpilibsuite/allwpilib/issues/1436

wpilib/tests/test_driverstation.py
```diff
@@ -125,8 +125,6 @@ def test_getStickPOV_limits(dsmock, halmock):
         dsmock.getStickPOV(-1, 1)
     with pytest.raises(IndexError):
         dsmock.getStickPOV(dsmock.kJoystickPorts, 1)
-    with pytest.raises(IndexError):
-        dsmock.getStickPOV(0, -1)
```
auscompgeek: Did you want to re-add this?

wpilib/tests/test_driverstation.py
```diff
@@ -109,10 +109,10 @@ def test_getStickAxis_limits(dsmock, halmock):
         dsmock.getStickAxis(-1, 1)
     with pytest.raises(IndexError):
         dsmock.getStickAxis(dsmock.kJoystickPorts, 1)
-    with pytest.raises(IndexError):
-        dsmock.getStickAxis(0, -1)
     with pytest.raises(IndexError):
         dsmock.getStickAxis(0, halmock.kMaxJoystickAxes)
+    with pytest.raises(IndexError):
+        dsmock.getStickAxis(0, -1)
```
auscompgeek: Not entirely sure why this call was moved.

wpilib/wpilib/driverstation.py
```diff
@@ -231,19 +323,22 @@ def getStickAxis(self, stick: int, axis: int) -> float:
 
         :returns: The value of the axis on the joystick.
         """
-        if axis < 0 or axis >= hal.kMaxJoystickAxes:
+        if 0 > axis or axis >= hal.kMaxJoystickAxes:
```
auscompgeek: I too love changing random conditions to Yoda comparisons.

wpilib/wpilib/driverstation.py
```diff
@@ -231,19 +323,22 @@ def getStickAxis(self, stick: int, axis: int) -> float:
...
 
         with self.cacheDataMutex:
             joystickAxes = self.joystickAxes[stick]
 
-            if axis >= joystickAxes.count:
-                self._reportJoystickUnpluggedWarning("Joystick axis %d on port %d not available, check if controller is plugged in\n" % (axis, stick))
-                return 0.0
+            axis_unavailable = axis >= joystickAxes.count
+
+            if not axis_unavailable:
```
auscompgeek: Seems redundant. This line is nowhere close to being long.

wpilib/wpilib/driverstation.py
```diff
@@ -254,18 +349,18 @@ def getStickPOV(self, stick: int, pov: int) -> int:
...
             joystickPOVs = self.joystickPOVs[stick]
+            pov_not_available = pov >= joystickPOVs.count
 
-            if pov >= joystickPOVs.count:
-                self._reportJoystickUnpluggedWarning("Joystick POV %d on port %d not available, check if controller is plugged in\n" % (pov, stick))
-                return -1
-            return joystickPOVs.povs[pov]
+        if pov_not_available:
+            self._reportJoystickUnpluggedWarning("Joystick POV %d on port %d not available, check if controller is plugged in\n" % (pov, stick))
+        return joystickPOVs.povs[pov]
```
auscompgeek: This seems to no longer return `-1` in the unavailable case. Is this intentional? I don't think this is what upstream does.

ariovistus: it is what upstream does

wpilib/wpilib/driverstation.py
```diff
@@ -388,12 +424,11 @@ def getJoystickIsXbox(self, stick: int) -> bool:
...
             raise IndexError("Joystick index is out of range, should be 0-%s" % self.kJoystickPorts)
 
-        with self.cacheDataMutex:
-            # TODO: Remove this when calling for descriptor on empty stick no longer crashes.
-            if 1 > self.joystickButtons[stick].count and 1 > len(self.joystickAxes[stick]):
-                self._reportJoystickUnpluggedWarning("WARNING: Joystick on port {} not avaliable, check if controller is "
-                                                     "plugged in.\n".format(stick))
-                return False
+        # TODO: Remove this when calling for descriptor on empty stick no longer crashes.
+        if 1 > self.joystickButtons[stick].count and 1 > len(self.joystickAxes[stick]):
```
virtuald: Neither of these checks are present upstream anymore. Presumably it doesn't crash anymore?

wpilib/wpilib/driverstation.py
virtuald: Yeah, it's not necessary here.

PR #511, opened 2018-09-30
update iterativerobot for 2019

ariovistus:


wpilib/wpilib/watchdog.py
```diff
@@ -0,0 +1,83 @@
+# validated: 2018-09-30 EN a818c7fd4741 edu/wpi/first/wpilibj/Watchdog.java
+# ----------------------------------------------------------------------------
+# Copyright (c) 2018 FIRST. All Rights Reserved.
+# Open Source Software - may be modified and shared by FRC teams. The code
+# must be accompanied by the FIRST BSD license file in the root directory of
+# the project.
+# ----------------------------------------------------------------------------
+
+import wpilib
```
auscompgeek: Erm.

wpilib/wpilib/watchdog.py
```diff
@@ -0,0 +1,83 @@
...
+        self.epochs.clear()
+        self.notifier.startPeriodic(self.timeout)
+
+    def disable(self):
+        """Disable the watchdog."""
+        self.notifier.stop()
+
+    def timeoutFunc(self):
+        if not self._isExpired:
+            print ("Watchdog not fed after %ss" % (m_timeout,))
```
auscompgeek: Don't call `print`, use the `logging` module instead.
```python
logger = logging.getLogger(__name__)

logger.info("blah %s", "asdf")
```

wpilib/wpilib/watchdog.py
```diff
@@ -0,0 +1,83 @@
...
+        return self._isExpired
+
+    def addEpoch(self, epochName):
+        """Adds time since last epoch to the list printed by printEpochs().
+
+        :param epochName: The name to associate with the epoch.
+        """
+        currentTime = Timer.getFPGATimestamp()
+        self.epochs[epochName] = currentTime - self.startTime
+        self.startTime = currentTime;
```
auscompgeek: semicolon?

wpilib/wpilib/watchdog.py
```diff
@@ -0,0 +1,83 @@
...
+
+class Watchdog:
+    """A class that's a wrapper around a watchdog timer.
+
+    When the timer expires, a message is printed to the console and an optional user-provided
+    callback is invoked.
+
+    The watchdog is initialized disabled, so the user needs to call enable() before use.
+
+    """
```
auscompgeek: Why is there a blank line before the end quotes here?

wpilib/wpilib/iterativerobotbase.py
```diff
@@ -60,8 +61,14 @@ class Mode(enum.Enum):
 
     # ----------- Overridable initialization code -----------------
 
-    def __init__(self):
+    def __init__(self, period):
```
auscompgeek: Seeing as we want to type hint everything, mind adding type hints here?

wpilib/wpilib/watchdog.py
```diff
@@ -0,0 +1,83 @@
...
+        self.notifier = wpilib.notifier.Notifier(self.timeoutFunc)
+        self.enable()
+
+    def getTime(self):
+        """Get the time in seconds since the watchdog was last fed."""
+        return Timer.getFPGATimestamp() - self.startTime
+
+    def isExpired(self):
+        """Returns true if the watchdog timer has expired."""
+        return self._isExpired
```
auscompgeek: Where is `self._isExpired` initialised?

wpilib/wpilib/iterativerobotbase.py
```diff
@@ -60,8 +61,14 @@ class Mode(enum.Enum):
...
+        """Constructor for IterativeRobotBase.
+
+        :param period: Period in seconds
+        """
         self.last_mode = self.Mode.kNone
+        self.period = period
+        self.watchdog = Watchdog(period, self.printLoopOverrunMessage)
         super().__init__()
```
auscompgeek: Hm. I wonder why this super call is at the end rather than the beginning. Realistically it should be before anything creates any other WPILib objects. Mind moving this?

wpilib/wpilib/iterativerobotbase.py
```diff
@@ -148,34 +155,52 @@ def testPeriodic(self):
     def loopFunc(self):
         """Call the appropriate function depending upon the current robot mode"""
```
auscompgeek: nit: There shouldn't be a blank line between a method docstring and the method body.

wpilib/tests/test_watchdog.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+from unittest.mock import MagicMock
+
+
+def test_init1(wpilib, sim_hooks, MockNotifier):
+    sim_hooks.time = 4.5
+    doggo = wpilib.Watchdog(2)
```
twinters007: I like it, but not sure how @virtuald feels about the 'informal' nature of this wording

wpilib/wpilib/watchdog.py
```diff
@@ -0,0 +1,83 @@
...
+        :param epochName: The name to associate with the epoch.
+        """
+        currentTime = Timer.getFPGATimestamp()
+        self.epochs[epochName] = currentTime - self.startTime
+        self.startTime = currentTime;
+
+    def printEpochs(self):
+        """Prints list of epochs added so far and their times."""
+        for key, value in self.epochs.items():
+            print ("\t%s: %ss" % (key, value))
```
twinters007: Same here. Don't print

wpilib/wpilib/watchdog.py
ariovistus: the usual way seemed to preclude mocking notifier - by the time you call patch, the original value is already in watchdog's namespace

wpilib/wpilib/watchdog.py
ariovistus: its what convert_javadoc.html outputs. must have missed that one

wpilib/wpilib/watchdog.py
auscompgeek: Not necessarily.

You might also want to consider mocking the Notifier methods instead of the entire class.

wpilib/tests/test_watchdog.py
auscompgeek: It's in the tests, who cares :stuck_out_tongue:

wpilib/wpilib/timedrobot.py
```diff
@@ -25,17 +25,15 @@ class TimedRobot(IterativeRobotBase):
...
-    DEFAULT_PERIOD = .02
+    kDefaultPeriod = .02
 
 
-    def __init__(self):
-        super().__init__()
-        hal.report(hal.UsageReporting.kResourceType_Framework, hal.UsageReporting.kFramework_Iterative)
+    def __init__(self, period: float=None):
+        if period is None:
+            period = TimedRobot.kDefaultPeriod
```
auscompgeek: I think the default should be in the method signature here. Would you agree @virtuald?

wpilib/wpilib/watchdog.py
```diff
@@ -0,0 +1,87 @@
...
+        """Disable the watchdog."""
+        self.notifier.stop()
+
+    def timeoutFunc(self) -> None:
+        if not self._isExpired:
+            logger.info("Watchdog not fed after %ss" % (self.timeout,))
+            self.callback()
+            self._isExpired = True
+            self.disable()
+
```
auscompgeek: nit: no blank line at EOF

```suggestion
```

wpilib/wpilib/timedrobot.py
virtuald: I agree.

wpilib/wpilib/watchdog.py
virtuald: Don't waste your time with nits, once we switch to an autoformatted it'll either be right or it'll be wrong. :)

PR #515, opened 2018-10-21
Really update UsageReporting for real this time

auscompgeek:


auscompgeek:
Ugh. Why is this is in #491 instead of the actual PR that updated these.

ariovistus:
I was just following your suggestion there

PR #517, opened 2018-10-31
validate namespace change

ariovistus:


auscompgeek:
hm?

ariovistus:
a bunch of stuff like

+import edu.wpi.first.hal.FRCNetComm.tInstances;
+import edu.wpi.first.hal.FRCNetComm.tResourceType;
+import edu.wpi.first.hal.HAL;
 import edu.wpi.first.networktables.NetworkTableEntry;
-import edu.wpi.first.wpilibj.hal.FRCNetComm.tInstances;
-import edu.wpi.first.wpilibj.hal.FRCNetComm.tResourceType;
-import edu.wpi.first.wpilibj.hal.HAL;

auscompgeek:
This seems like a silly way of handling that though. After all, it's a single commit that doesn't result in any changes on our end, right?

ariovistus:
Don't know, I haven't gone wading through upstream commits. If it is the case that all these are happening in one commit, what's the right way of handling this?

virtuald:
If they haven't changed any code behavior, then you can add that commit to the exclude commits file.

PR #519, opened 2018-11-03
update trigger for 2019

ariovistus:


ariovistus:
whileHeld seems to be proxying whileActive

auscompgeek:
Ah right, Button has a few more-meaningful aliases for those methods. Good catch.

PR #521, opened 2018-11-13
update differentialdrive for 2019

ariovistus:


wpilib/wpilib/drive/differentialdrive.py
```diff
@@ -118,24 +118,25 @@ def __init__(self, leftMotor, rightMotor):
         self.quickStopAlpha = self.kDefaultQuickStopAlpha
         self.quickStopAccumulator = 0.0
         self.reported = False
+        self.rightSideInvertMultiplier = -1.0
```
auscompgeek: We could probably just store this as an int. But it's fine as is.

PR #523, opened 2018-11-17
update timer for 2019

ariovistus:
also validate a few others


wpilib/wpilib/analogtriggeroutput.py
```diff
@@ -1,4 +1,4 @@
-# validated: 2017-12-27 TW f9bece2ffbf7 edu/wpi/first/wpilibj/AnalogTriggerOutput.java
+# validated: 2018-11-17 EN 0e9172f9a708 edu/wpi/first/wpilibj/AnalogTriggerOutput.java
```
auscompgeek: That commit seems like something we might want to exclude. Mind looking into it?

ariovistus: synchronized was added to ConditionalCommand._cancel, but the rest looks ignorable

PR #525, opened 2018-11-18
port shuffleboard support for 2019

ariovistus:


wpilib/wpilib/iterativerobot.py
```diff
@@ -21,6 +22,8 @@ class IterativeRobot(IterativeRobotBase):
 
     periodic() functions from the base class are called each time a new packet is received from
     the driver station.
+
+    .. deprecated:: 2019.0.0
```
auscompgeek: Should add the deprecation reason in the docs too.

wpilib/wpilib/iterativerobot.py
```diff
@@ -41,6 +44,10 @@ def __init__(self) -> None:
                      the communications code starts.
         """
         super().__init__(IterativeRobot.kPacketPeriod)
+        warnings.warn(
+            "IterativeRobot is deprecated. Use TimedRobot instead. It's a " +
+            "drop-in replacement that provides more regular execution periods."
+        )
```
auscompgeek: `category=DeprecationWarning`?

wpilib/wpilib/shuffleboard/shuffleboard.py
```diff
@@ -0,0 +1,112 @@
...
+            .withWidget("Toggle Button")
+            .getEntry())
+
+
+    Changing the colors of the boolean box:
+    .. code-block:: python
+        myBoolean = (Shuffleboard.getTab("Example Tab")
+            .add("My Boolean", False)
+            .withWidget("Boolean Box")
+            # TODO: WATWATWAT
```
virtuald: Haha. Please fix the example... should be `{'colorWhenTrue': 'green', .. }`

wpilib/wpilib/shuffleboard/widget.py
```diff
@@ -0,0 +1,23 @@
+# validated: 2018-11-18 EN 175c6c1f0130 edu/wpi/first/wpilibj/shuffleboard/ShuffleboardWidget.java
+from .component import ShuffleboardComponent
+from .container import ShuffleboardContainer
+
+class ShuffleboardWidget(ShuffleboardComponent):
+    """
+    Abstract superclass for widgets.
+
+    This class is package-private to minimize API surface area.
```
virtuald: Remove this comment?

wpilib/wpilib/shuffleboard/container.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        Adds a widget to this container to display the given sendable.
+
+        raises AssertionError if a widget already exists in this container 
+        with the given title
+
+        :param title:    the title of the widget. if not provided, `sendable.getName()` is used.
+        :param sendable: the sendable to display
+        :param defaultValue: the default value of the widget. not to be combined with `sendable`
+        :returns: a widget to display the sendable data
```
virtuald: Should add a python-specific note that empty lists are not usable here.

wpilib/wpilib/shuffleboard/shuffleboard.py
auscompgeek: I guess things like this are why we recommend that people that wish to help port changes actually know Java. :)

wpilib/wpilib/shuffleboard/component.py
```diff
@@ -0,0 +1,109 @@
...
+    def getType(self) -> str:
+        return self.type
+
+    def getTitle(self) -> str:
+        return self.title
+
+    def getProperties(self):
+        return self.properties
+
+    def withProperties(self, properties: dict) -> 'ShuffleboardComponent':
```
auscompgeek: Are there particular types the keys and values need to be? If so, that should also be noted in the type hint.

wpilib/wpilib/shuffleboard/container.py
ariovistus: in the default value?

ariovistus: why not?

virtuald: Because it wouldn't be possible to know the type of the list.

ariovistus: that's kind of an annoying restriction. would it make sense to have networktables accept something like typing.List[str] to signal you want an empty list of strings?

auscompgeek: There's no way to uniquely represent an empty value of that type ­— the type of the elements could be anything.

virtuald: In theory we could make simple wrappers around list:

```
class StringList(list):
    pass
```

... but that seems more trouble than it's worth. In python, users can just use putXXXArray if they absolutely must put an empty array of things in.

PR #526, opened 2018-11-22
fix typos in killough drive

ariovistus:


wpilib/tests/test_drive.py
```diff
@@ -379,3 +461,25 @@ def test_stopMotor_4(drive_mecanum):
 
 def test_feedWatchdog(drive_diff):
     drive_diff.feedWatchdog()
+
+
+def test_killough_driveCartesian(wpilib):
+    m1, m2, m3 = MagicMock(), MagicMock(), MagicMock()
+    drive = wpilib.drive.KilloughDrive(m1, m2, m3)
+
+    drive.driveCartesian(0.5, 0.5, 0.5, 90)
```
auscompgeek: It's not immediately clear how these values correspond to the output values below. It might be worth having some simpler input values so someone sanity-checking these tests understand where the values come from.

"Testing" an implementation by plugging in values to the implementation and taking its output values is not a great unit test - that's merely a regression test.

auscompgeek: Fair. I'm mostly concerned about how brittle tests like these can end up being in the long run.

Regardless, this PR does fix a typo, so we should probably merge this as it is, and adjust the tests at a later point.

auscompgeek: Speaking of which, I'm surprised we don't have example robot code that exercise these...

PR #528, opened 2018-12-02
Format code with black

virtuald:
The time has come.


docs/gensidebar.py
```diff
@@ -5,86 +5,91 @@
...
+    lines = ["", ".. DO NOT MODIFY! THIS PAGE IS AUTOGENERATED!", ""]
+
     def toctree(name):
-        lines.extend(['.. toctree::',
-                      '    :caption: %s' % name,
-                      '    :maxdepth: 2',
-                      ''])
-    
+        lines.extend(
+            [".. toctree::", "    :caption: %s" % name, "    :maxdepth: 2", ""]
```
ariovistus: does black require these to be on one line? I think it looks better when each item is on its own line

virtuald: Black isn't really configurable, except by turning formatting off. I agree this part is ugly, but it's not really user-facing so I'm fine with that.

wpilib/tests/test_drive.py
```diff
@@ -248,16 +259,21 @@ def check_curvature(wpimock, drive_diff, y, rotation, isQuickTurn):
...
-    drive_diff.rightMotor.set.assert_called_once_with(-rightMotorSpeed * drive_diff.maxOutput)
+    drive_diff.leftMotor.set.assert_called_once_with(
+        leftMotorSpeed * drive_diff.maxOutput
+    )
+    drive_diff.rightMotor.set.assert_called_once_with(
+        -rightMotorSpeed * drive_diff.maxOutput
+    )
 
 
+# fmt: off
```
ariovistus: why is this one needed?

docs/gensidebar.py
auscompgeek: The alternative is to call `lines.append` a bunch of times, I guess. Follow your heart, I suppose.

wpilib/tests/test_drive.py
ariovistus: I know. It looks like the code you're ignoring is the test cases on 280 - 289. any particular reason they are 4 to a line?

virtuald: Not really, but the formatting black applies is much worse IIRC.

wpilib/tests/test_pwm_motor_controllers.py
```diff
@@ -5,40 +5,53 @@
...
+)
 def test_controller(wpilib, hal_data, clsname):
-    
+
     # create object/helper function
     obj = getattr(wpilib, clsname)(2)
-    
+
     # validate reporting is correct
     # TODO: remove the VictorSPX and DMC60 check once there is usage reporting
```
ariovistus: out of scope for this pr, but these two have usage reporting now

wpilib/tests/test_shuffleboard.py
```diff
@@ -23,7 +23,7 @@ def test_addPersistent(wpilib, getTable):
     tab = wpilib.shuffleboard.Shuffleboard.getTab("tacos")
     widget = tab.addPersistent("simplevalue", 1)
     # k, how to kick networktables?
-    #assert widget.getEntry().isPersistent()
+    # assert widget.getEntry().isPersistent()
```
ariovistus: rats, did I forget to ask about this for the shuffleboard pr?

auscompgeek: lol. that should be a thing, unless you forgot to make `getEntry` a thing.

hal-roborio/hal_impl/types.py
```diff
@@ -1,3 +1,5 @@
+# fmt: off
```
auscompgeek: Surely this doesn't look _that_ terrible after formatting.

virtuald: It's not very good, as it's fairly abnormal code. And I'd rather it stay the way it is.

virtuald:
All your pull request are belong to me.

virtuald:
@ariovistus you'll need to rebase all your stuff. Sorry, but not sorry? In the long run, you'll thank me because @auscompgeek will find less things to nitpick about ;)

auscompgeek:
yay I finally no longer have to care about style :stuck_out_tongue_winking_eye:

PR #534, opened 2018-12-12
HAL updates for 2019

virtuald:
Didn't fix any of the underlying stuff, I'm sure the tests will fail.

### DIFF
--- a/wpilib/tests/spec_scanners/hal_scanner.py
+++ b/wpilib/tests/spec_scanners/hal_scanner.py
@@ -368,6 +368,9 @@ class Class:
                 elif hasattr(o, "fndata"):
                     self.add_function(Function.from_py(name, o))
 
+                elif o is NotImplemented:
+                    self.add_not_implemented_function(name)
+
         return self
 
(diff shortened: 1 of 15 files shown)


wpilib/tests/spec_scanners/hal_scanner.py
```diff
@@ -68,6 +69,33 @@ def get_hal_dirs(hal_dir):
...
+        "--passthru-unfound-includes",
+        "-I",
+        dirname(dirname(fname)),
+        fname,
+        "-o",
+        ppname,
+    ]
+    subprocess.check_call(args)
+
+    return CppHeaderParser.CppHeader(ppname)
```
auscompgeek: Wouldn't it be better to have pcpp dump to stdout and feed it in here with `argType="string"`?

virtuald: No, because it breaks and I didn't feel like digging into it.

virtuald:
Adding PDP and SPI updates to this.

PR #535, opened 2018-12-12
Fix PIDController test to work with pytest 4+

twinters007:


ariovistus:
it will, i'll rebase, not a problem.

haven't used pytest 4, how was it not working before this fix?

twinters007:
@ariovistus by calling `pid(...)` you're directly calling a fixture, and pytest got fed up with it

PR #540, opened 2018-12-18
Update ultrasonic

ArchdukeTim:


wpilib/wpilib/ultrasonic.py
```diff
@@ -59,6 +59,8 @@ class Unit:
     #: ultrasonic sensor list
     sensors = weakref.WeakSet()
 
+    _sensor_mutex = threading.Lock()
```
auscompgeek: This name is confusing. This only guards the `sensors` variable. Shouldn't it be named after said variable?

twinters007:
The code that was there doesn't match the code that the verification commit matches, but it does the same thing as the current code in wpilibj (more or less). It loops through the encoders, increments a counter, pings, then returns if the counter is 0. I guess I could get rid of the None check and the automatic check. I'm not sure why it is this way, but I imagine it was done differently for a reason. If not, then I can change it

auscompgeek:
I'd imagine it would've been done this way because whoever ported it originally didn't want to have an invasive linked list.

auscompgeek:
@virtuald Upstream have added synchronisation around the sensor list whilst changing from a linked list to a container, but the critical section doesn't contain the entire loop. I don't see a good way of doing that whilst still using a WeakSet. Thoughts?

virtuald:
I think I had assumed the WeakSet was already thread safe. Probably should verify that, and then put in a note about the python-specific changes. Agree that the mutex isn't ideal.

virtuald:
WeakSet is not threadsafe, so the mutex would be needed. However, it seems that the mutex upstream doesn't cover all uses of the set, so we need to probably file a bug for that (and fix our own usage).

auscompgeek:
I think the best way to actually do this is probably to make a local copy of the sensor set, and do everything else outside the critical section.

Taking another look at the differences here... it seems that we have a mutex around `automaticEnabled` where upstream doesn't (and it doesn't seem to protect anything else), and there might be some bugs we've introduced too.

auscompgeek:
@ArchdukeTim what's the status on this?

virtuald:
OBE

ArchdukeTim:
I like how we got through a whole season without needing this

auscompgeek:
Clearly nobody has ever actually used Ultrasonic.

PR #541, opened 2018-12-19
hal-roborio: Use union to get ControlWord bits

auscompgeek:
This should be faster (and saner!) than the type punting shenanigans I was doing earlier.

### DIFF
--- a/hal-roborio/hal_impl/types.py
+++ b/hal-roborio/hal_impl/types.py
@@ -34,7 +34,7 @@ __all__ = [
 # HAL
 #############################################################################
 
-class ControlWord(C.Structure):
+class _ControlWord(C.Structure):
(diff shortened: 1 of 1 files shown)


auscompgeek:
(Un)scientific comparison on an x86-64 machine:

```
In [5]: c = _ControlWord(1, 0, 0, 0, 1, 1)

In [6]: %timeit c.to_bits()  # old implementation                 
2.41 µs ± 80.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [7]: c = ControlWord(c)                      

In [8]: %timeit c.bits                   
61.4 ns ± 1.32 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [11]: assert c._.to_bits() == c.bits                                                                   
```

I'd imagine the speedup magnitude would be the same.

PR #542, opened 2018-12-19
MotorSafety and InterruptableSensorBase updates for 2019

ArchdukeTim:


auscompgeek:
We should split out the InterruptableSensorBase change, because that looks good and shouldn't get held up by the broken-in-tests MotorSafety. @nickschatz

nickschatz:
I reverted motorsafety.py, leaving this commit with no working motor safety because I left the DriverStation.py updates which removes the checkmotors thread. I can split that out as well if you'd rather.

What's the approach for fixing MotorSafety in tests?

auscompgeek:
Similar way to how DriverStation, Notifier, PIDController and Watchdog itself are tested.

Also yeah don't split that commit.

nickschatz:
@auscompgeek please review. Only changes should be updating InterruptibleSensorBase.

PR #544, opened 2018-12-19
Remove SensorBase

auscompgeek:


ariovistus:
https://github.com/robotpy/robotpy-navx/blob/master/navx/ahrs.py#L36

PR #546, opened 2018-12-25
Update Watchdog to use a single thread

auscompgeek:
Also fix a couple of logging-related format string bugs.

### DIFF
--- a/wpilib/wpilib/watchdog.py
+++ b/wpilib/wpilib/watchdog.py
@@ -240,7 +240,10 @@ class Watchdog:
                                     watchdog._timeout / 1e6,
                                 )
                         lock.release()
-                        watchdog._callback()
+                        try:
+                            watchdog._callback()
+                        except Exception:
+                            logger.exception("Uncaught exception in Watchdog callback")
                         lock.acquire()
                         watchdog._isExpired = True
 
--- a/wpilib/wpilib/watchdog.py
+++ b/wpilib/wpilib/watchdog.py
@@ -11,7 +11,6 @@ import threading
 from typing import Callable
 
 import hal
-from .resource import Resource
 
 import logging
 
@@ -33,7 +32,7 @@ class Watchdog:
     # Used for timeout print rate-limiting
     kMinPrintPeriod = 1000000  # us
 
-    _watchdogs = []
+    _watchdogs = []  # type: List[Watchdog]
     _queueMutex = threading.Lock()
     _schedulerWaiter = threading.Condition(_queueMutex)
     _keepAlive = True
@@ -41,14 +40,15 @@ class Watchdog:
 
     @classmethod
     def _reset(cls) -> None:
-        thread = cls._thread
+        with cls._queueMutex:
+            thread = cls._thread
         if thread is not None:
             cls._keepAlive = False
             with cls._queueMutex:
                 cls._watchdogs.clear()
                 cls._schedulerWaiter.notify_all()
+                cls._thread = None
             thread.join()
-        cls._thread = None
         cls._keepAlive = True
 
     def __init__(self, timeout: float, callback: Callable[[], None]) -> None:
@@ -72,11 +72,12 @@ class Watchdog:
         #: prints a more specific message.
         self.suppressTimeoutMessage = False
 
-        Resource._add_global_resource(self)
-
-        if Watchdog._thread is None:
-            Watchdog._thread = threading.Thread(target=self._schedulerFunc, daemon=True)
-            Watchdog._thread.start()
+        with self._queueMutex:
+            if Watchdog._thread is None:
+                Watchdog._thread = threading.Thread(
+                    target=Watchdog._schedulerFunc, daemon=True
+                )
+                Watchdog._thread.start()
 
     # Elements with sooner expiration times are sorted as lesser.
     # Python's heap queue is a min-heap.
@@ -96,9 +97,6 @@ class Watchdog:
         if self.__class__ is not other.__class__:
             return NotImplemented
         return self._expirationTime > other._expirationTime
-
-    def close(self) -> None:
-        self.disable()
 
     def getTime(self) -> float:
         """Returns the time in seconds since the watchdog was last fed."""
--- a/wpilib/wpilib/watchdog.py
+++ b/wpilib/wpilib/watchdog.py
@@ -41,11 +41,11 @@ class Watchdog:
 
     @classmethod
     def _reset(cls) -> None:
-        cls._watchdogs = []
         thread = cls._thread
         if thread is not None:
             cls._keepAlive = False
             with cls._queueMutex:
+                cls._watchdogs.clear()
                 cls._schedulerWaiter.notify_all()
             thread.join()
         cls._thread = None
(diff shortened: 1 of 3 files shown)


auscompgeek:
Wait, we should probably port the tests across as well.

auscompgeek:
Hmm. The upstream tests sleep to wait for timeouts. My attempts to replicate the tests have been unsuccessful...

wpilib/wpilib/watchdog.py
```diff
@@ -75,18 +171,73 @@ def reset(self) -> None:
...
-        self.epochs.clear()
-        self.notifier.startPeriodic(self.timeout)
+        self._epochs.clear()
+
+        with self._queueMutex:
+            self._isExpired = False
+
+            watchdogs = self._watchdogs
+            try:
+                watchdogs.remove(self)
```
virtuald: `watchdogs.pop(self, None)`, no exception handling necessary.

wpilib/wpilib/watchdog.py
```diff
@@ -26,45 +30,137 @@ class Watchdog:
...
+        self._epochs.clear()
+        timeout = int(timeout * 1e6)
+
+        with self._queueMutex:
+            self._timeout = timeout
+            self._isExpired = False
+
+            watchdogs = self._watchdogs
+            try:
+                watchdogs.remove(self)
```
virtuald: See other comment.

wpilib/wpilib/watchdog.py
```diff
@@ -75,18 +171,73 @@ def reset(self) -> None:
...
-            self.callback()
-            self._isExpired = True
-            self.disable()
+        """Disables the watchdog timer."""
+        with self._queueMutex:
+            self._isExpired = False
+
+            watchdogs = self._watchdogs
+            try:
+                watchdogs.remove(self)
```
virtuald: Ditto.

wpilib/wpilib/watchdog.py
```diff
@@ -75,18 +171,73 @@ def reset(self) -> None:
...
+                        watchdog._callback()
+                        cls._queueMutex.acquire()
+                        watchdog._isExpired = True
+
+                    # Otherwise, a Watchdog removed itself from the queue (it
+                    # notifies the scheduler of this), we are being reset (by
+                    # the test harness) or a spurious wakeup occurred, so rewait
+                    # with the soonest watchdog timeout.
+                else:
+                    while cls._keepAlive and not cls._watchdogs:
```
virtuald: Add a note that this is python specific.

wpilib/wpilib/watchdog.py
```diff
@@ -75,18 +171,73 @@ def reset(self) -> None:
...
+
+    @classmethod
+    def _schedulerFunc(cls) -> None:
+        getFPGATime = hal.getFPGATime
+
+        with cls._queueMutex:
+            while cls._keepAlive:
+                if cls._watchdogs:
+                    delta = cls._watchdogs[0]._expirationTime - getFPGATime()
+                    timedOut = cls._schedulerWaiter.wait(delta)
```
virtuald: Condition variables return False when timed out, maybe that's why you weren't able to get the tests to work?

wpilib/wpilib/watchdog.py
```diff
@@ -75,18 +171,73 @@ def reset(self) -> None:
...
+                        # If the condition variable timed out, that means a Watchdog
+                        # timeout has occurred, so call its timeout function.
+                        watchdog = cls._watchdogs[0]
+
+                        if not watchdog.suppressTimeoutMessage:
+                            logger.info(
+                                "Watchdog not fed after %.6fs", watchdog._timeout / 1e6
+                            )
+                        cls._queueMutex.release()
+                        watchdog._callback()
```
virtuald: Hm. I feel like we should catch exceptions + log them here instead of letting the world burn down if a callback throws.

wpilib/wpilib/watchdog.py
```diff
@@ -75,18 +171,73 @@ def reset(self) -> None:
...
+                        ):
+                            continue
+
+                        # If the condition variable timed out, that means a Watchdog
+                        # timeout has occurred, so call its timeout function.
+                        watchdog = cls._watchdogs[0]
+
+                        if not watchdog.suppressTimeoutMessage:
+                            logger.info(
+                                "Watchdog not fed after %.6fs", watchdog._timeout / 1e6
```
virtuald: Heh, watchdogs don't have names? That seems like it would be useful...

wpilib/wpilib/watchdog.py
virtuald: Oh, I thought it was a dict.

wpilib/wpilib/watchdog.py
virtuald: With java you would need to throw an unchecked exception otherwise you'll get a compile error. I feel this is one of the cases where we need to differ from them, as it's significantly easier to get an exception for Python.

auscompgeek: This is a scenario where you could potentially end up in an infinite loop of exceptions if we do decide to catch them though.

virtuald: Remove the offending watchdog if the exception is thrown?

auscompgeek: Hm, maybe. Code might get rather messy though.

How many teams would you expect to create their own Watchdog objects anyway?

virtuald: I'm more worried that some user setting causes our code to do something bad?, and they can't easily figure out what it is. Seems bad to kill all the watchdogs when it's just one bad apple.

auscompgeek:
I would appreciate it when it comes time to merge this that we squash this.

wpilib/wpilib/watchdog.py
```diff
@@ -205,43 +205,49 @@ def disable(self) -> None:
...
                             watchdog._lastTimeoutPrintTime = now
                             if not watchdog.suppressTimeoutMessage:
-                                logger.info(
+                                log(
                                     "Watchdog not fed after %.6fs",
                                     watchdog._timeout / 1e6,
                                 )
-                        cls._queueMutex.release()
+                        lock.release()
                         watchdog._callback()
```
virtuald: At the very least, if you're not going to catch the exception, use try...finally to release the lock.

auscompgeek: You seem to be talking to an outdated version of this PR.

wpilib/wpilib/watchdog.py
```diff
@@ -26,45 +30,137 @@ class Watchdog:
...
-        self.notifier = Notifier(self.timeoutFunc)
-        self.enable()
+
+        #: Enable or disable suppression of the generic timeout message.
+        #:
+        #: This may be desirable if the user-provided callback already
+        #: prints a more specific message.
+        self.suppressTimeoutMessage = False
+
+        Resource._add_global_resource(self)
```
auscompgeek: Hmm. If we have the `_reset` above, is this necessary? @virtuald

auscompgeek: I've gone ahead and removed this and the corresponding redundant `close` method then.

PR #547, opened 2018-12-25
Silence DeprecationWarning in RobotDrive tests

auscompgeek:


wpilib/tests/test_robotdrive.py
```diff
@@ -2,6 +2,8 @@
 from unittest.mock import MagicMock
 import math
 
+pytestmark = pytest.mark.filterwarnings("ignore:RobotDrive :DeprecationWarning")
```
ArchdukeTim: Why are you saving it as a variable?

auscompgeek: https://docs.pytest.org/en/latest/warnings.html#pytest-mark-filterwarnings

PR #549, opened 2018-12-26
wpilib: Fix DeprecationWarnings

auscompgeek:


wpilib/wpilib/robotdrive.py
```diff
@@ -22,7 +22,10 @@ def _freeRobotDrive(allocatedSpeedControllers):
     Free the speed controllers if they were allocated locally
     """
     for sc in allocatedSpeedControllers:
-        if hasattr(sc, "free"):
+        sc_close = getattr(sc, "close", None)
```
ArchdukeTim: Why make updates to a deprecated class? Is it just to silence the warnings in pytest? It would probably make more sense to do it on the testing side, since this introspection is (more or less) unique to python, and this wouldn't show up in wpilibj.

Maybe you or Dustin have better arguments for this, but it seems silly to come back to this and try and make it work with new code.

auscompgeek: A user can pass in a motorController keyword argument and have RobotDrive allocate the motor controller object for them. If that class has a close method but didn't have a free method, that resource could leak.

Correctness over ideals.

ArchdukeTim: Fair point

PR #553, opened 2019-01-02
shuffleboard: Fix and clean up

auscompgeek:
Fix a couple of bugs introduced by us:
 - Adding a widget with position column=0 or row=0 was not allowed.
 - A properties subtable was always created, even with no properties.
 - Checks that should always run were asserts.

Clean up unnecessary None checks. They only exist in Java because null
is easy to accidentally create.

Also clean up ShuffleboardContainer, in particular its add method to
avoid match_arglist - title is now a keyword-only argument instead.

Also fix the docs, ensuring that the public classes are reexported.


wpilib/wpilib/shuffleboard/component.py
```diff
@@ -13,8 +13,6 @@ class ShuffleboardComponent:
     def __init__(
         self, parent: ShuffleboardContainer, title: str, type: Optional[str] = None
     ):
-        assert parent is not None, "parent cannot be None"
```
virtuald: why remove this?

wpilib/wpilib/shuffleboard/container.py
```diff
@@ -42,83 +48,76 @@ def getLayout(self, type: str, title: str):
 
         return self.layouts[title]
 
-    def add(self, *args, **kwargs) -> "ShuffleboardWidget":
+    @overload
```
virtuald: ... heh

wpilib/wpilib/shuffleboard/component.py
auscompgeek: > Clean up unnecessary None checks. They only exist in Java because null is easy to accidentally create.

If the type signature is followed correctly, it should be obvious that you should never pass None in. To pass None, you'd have to explicitly state so.

wpilib/wpilib/shuffleboard/container.py
auscompgeek: Most appropriate way to document how one would get the different return types, I guess (even though the overload signatures don't show up in the docs). :man_shrugging:

PR #554, opened 2019-01-02
Use enums for constants

ArchdukeTim:


hal-base/hal/constants.py
```diff
@@ -1,3 +1,4 @@
+import enum
```
ArchdukeTim: Not sure if we wanna use enums for this class. It doesn't *hurt*, but I'm not sure if it helps now or will later

hal-base/hal/constants.py
```diff
@@ -7,12 +8,12 @@
...
 
 
-class RuntimeType:
+class RuntimeType(enum.IntEnum):
     Athena = 0
     Mock = 1
 
 
-class UsageReporting:
+class UsageReporting(enum.IntEnum):
```
auscompgeek: Given that this has a lot of duplicates with different semantics, I don't think this makes sense as a single enum.

ArchdukeTim: We probably don't need to. I don't think anything is type hinted as UsageReporting

virtuald:
Looks like it needs formatting.

hal-base/hal/constants.py
virtuald: Yeah, this just causes problems:

```
>>> hal.UsageReporting.kLanguage_LabVIEW
<UsageReporting.kResourceType_Module: 1>
```

PR #556, opened 2019-01-03
Hopefully final preseason bump

auscompgeek:


ariovistus:
you removed robotstate_impl in conftest.py. RobotState.isDisabled() needs to return false.

devtools/exclude_commits
```diff
@@ -19,3 +19,7 @@ e548a5f7054ccc221ede1917bb62f512d1fb0fe3 # PMD 6.3.0
 e2100730447d83b20267f7fff87fc1ca6bf47fb0 # Move HAL classes to their own base package  
 0e9172f9a70820f1a3ac8f5c8c55f63092e832d3 # Fix many errorprone warnings (#1247)
 acb786a7916621478166be85c32c20d9241c6c1c # Move MotorSafetyHelper into MotorSafety base class
+2e5fece594fd2098cdf83290370620d50b36f83a # utility class checks
+a60f312d19ee3a38df4318e1fec86c6789ef667a # eager null checks in drive base classes
```
virtuald: This seems like a good change to keep.

virtuald:
Working on this.

devtools/exclude_commits
auscompgeek: I dunno about this. Explicitly checking whether the user has passed in None when it is clearly documented that the function doesn't take None seems like a Python anti-pattern.

This was a Java-only change too. There's no corresponding C++ change.

virtuald: I agree -- but the point of the check is for the robot to fail immediately, instead of later on in the process. This seems like a Good Thing to me.

I haven't looked, but I'm guessing the C++ version takes references, which can never be null.

virtuald:
:shipit:

PR #558, opened 2019-01-04
Enable faulthandler

### LINKED ISSUE #266: Add faulthandler
For catching segfaults/other stuff

virtuald:
- This is incredibly useful for certain types of debugging, and AFAIK
  doesn't add any noticeable overhead, so we enable it by default
- Fixes #266

### DIFF
--- a/wpilib/wpilib/_impl/main.py
+++ b/wpilib/wpilib/_impl/main.py
@@ -2,6 +2,7 @@
 
 import argparse
 import inspect
+import os
 import sys
 
 from os.path import exists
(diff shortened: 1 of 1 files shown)


virtuald:
It would also be useful to add an entrypoint to pyfrc that could send SIGUSR2 to the process on the robot? Or add that to the vscode plugin? Or both perhaps.

auscompgeek:
Seems useful. Might cause some grief if teams were using SIGUSR2 for anything else, but seems quite unlikely.

PR #563, opened 2019-01-05
Small updates for 2019

ArchdukeTim:
I ended up just calling the setter myself, since the test is for SD/SendableChooser, and not really the ntcore functionality.


wpilib/wpilib/drive/differentialdrive.py
```diff
@@ -112,6 +112,8 @@ def __init__(self, leftMotor: SpeedController, rightMotor: SpeedController) -> N
         """
         super().__init__()
 
+        assert leftMotor is not None, "Left Motor should not be None"
```
ArchdukeTim: I think David actually added this check to exluded_commits so I'll get rid of these

virtuald: I put then back in.

virtuald:
Did you close this for a reason?

PR #564, opened 2019-01-05
Small Updates

ArchdukeTim:


virtuald:
Oh I see. You should rebase this.

wpilib/wpilib/serialport.py
```diff
@@ -201,12 +201,14 @@ def read(self, count: int) -> bytes:
         """
         return hal.readSerial(self.port, count)
 
-    def write(self, buffer: bytes) -> int:
+    def write(self, buffer: bytes, count: int) -> int:
```
virtuald: This is python, we don't need a count.

ArchdukeTim: I was questioning the java, but now I realize the JNI function needs the count

virtuald: Feel free to add a 'python-specific' note here so future reviewers aren't confused.

auscompgeek:
You broke the docs.

wpilib/wpilib/shuffleboard/widget.py
```diff
@@ -1,14 +1,21 @@
...
 # must be accompanied by the FIRST BSD license file in the root directory of
 # the project.
 # ----------------------------------------------------------------------------
+from typing import Union
+
 from .component import ShuffleboardComponent
 from .container import ShuffleboardContainer
 
 
+class WidgetType:
```
auscompgeek: Are you actually going to port the enums for this? Otherwise this is completely useless.

auscompgeek: Also I feel this should be an enum.Enum instead of a getter method to return its value.

ArchdukeTim: Why can't this just act as the interface class for type hints?

auscompgeek: I think we are again in violent agreement.

ArchdukeTim: But this works fine as is for the interface. Unless type hints has a way of specifying "idc what type it is, as long as it has these attributes"...should I change to `hasattr` within the method?

ArchdukeTim: oh I'm dumb. If they define it as one of the enums, it won't be a widget type...ugh ok

auscompgeek: Realistically, the only WidgetType that teams will probably care about is BuiltInWidgets. If users create their own widgets, they can pass in a string constant just fine.

ArchdukeTim: I've got a trick up my sleeve you'll probably dislike

auscompgeek: It'd probably be better to just have BuiltInWidgets by itself actually. More discoverable that way.

wpilib/wpilib/serialport.py
```diff
@@ -205,6 +205,7 @@ def write(self, buffer: bytes) -> int:
         """Write raw bytes to the serial port.
         
         :param buffer: The buffer of bytes to write.
+
         Python-Specific: No count parameter needed
```
auscompgeek: This really shouldn't be in the docstring, methinks.

virtuald: Yeah, it's an implementation note for future porters, users shouldn't care.

ArchdukeTim: ok. Below the docstring in the method, or class-level above the method?

virtuald:
This seems fine. Out of curiosity, have you tried the updated chooser with shuffleboard?

wpilib/wpilib/shuffleboard/widget.py
```diff
@@ -1,13 +1,28 @@
...
 # must be accompanied by the FIRST BSD license file in the root directory of
 # the project.
 # ----------------------------------------------------------------------------
+from typing import Union, TYPE_CHECKING
+
 from .component import ShuffleboardComponent
 from .container import ShuffleboardContainer
 
+if TYPE_CHECKING:
+    from typing_extensions import Protocol
```
auscompgeek: Pretty sure this will cause mypy errors if you don't have typing_extensions installed.

wpilib/wpilib/shuffleboard/widget.py
auscompgeek: @virtuald any strong opinions on the above?

wpilib/wpilib/shuffleboard/widget.py
ArchdukeTim: it will. Should we add it as an optional dep like you did with camera server?

wpilib/wpilib/serialport.py
virtuald: In the method below the docstring.

wpilib/wpilib/shuffleboard/widget.py
virtuald: If there's a base widget type that is almost always going to be used, I think it'd be better to use that for now instead of introducing a new dependency.

auscompgeek: Looking at the C++ implementation, all WidgetTypes have a string constant name. I say my enum idea earlier is probably the way to go.

wpilib/wpilib/shuffleboard/widget.py
virtuald: I think we decided to get rid of this? Is that right?

virtuald:
Thanks!

PR #566, opened 2019-01-07
'Implement' SpeedController

ArchdukeTim:


virtuald:
Why?

ArchdukeTim:
Type hints. SpeedControllerGroup accepts speedcontrollers, and stuff like `Spark` doesn't extend it, so the type check fails

virtuald:
Hm. From a typing perspective, shouldn't the various interfaces be protocols? None of them actually do anything.

ArchdukeTim:
I agree. I was waiting on your thoughts to either making them protocols or abstract base classes. If we make them protocols, we have to add the `typing_extension` module as a dep

virtuald:
I definitely don't want them to be abcs, but I also don't really want to add typing_extension as a dep that has to be installed on the robot either? Looking for a middle ground here...

virtuald:
OBE

PR #569, opened 2019-01-10
wpilib.shuffleboard: Fix builtin layouts/widgets

auscompgeek:
Open question: what table style do we prefer? Simple, CSV, or list? (not list pls)

Pandoc will output simple-style, but CSV might be able to automatically convert from the Javadocs :man_shrugging:


virtuald:
@auscompgeek if this is functional, let's merge it and push a tag. Sorry that I forgot about it.

PR #575, opened 2019-01-27
hal: Avoid circular imports

auscompgeek:


virtuald:
Hm, I suppose. Does this cause problems in `pip install -e` situations?

auscompgeek:
Worked fine in my venv with everything installed as editable (although I did do a reinstall just in case). But editable installs add the entire package root to sys.path anyway, so it shouldn't cause any problems.

PR #577, opened 2019-01-29
Warn users if they forgot to call wpilib.run

virtuald:
* Inspired by https://www.chiefdelphi.com/t/robot-code-not-deploying-from-robot-py/343672

### DIFF
--- a/wpilib/wpilib/_impl/main.py
+++ b/wpilib/wpilib/_impl/main.py
@@ -1,6 +1,7 @@
 # novalidate
 
 import argparse
+import atexit
 import inspect
 import os
 import sys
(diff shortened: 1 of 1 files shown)


wpilib/wpilib/_impl/main.py
```diff
@@ -11,6 +12,31 @@
 from .logconfig import configure_logging
 import hal_impl
 
+# Make silly mistakes more obvious
+_show_run_warning = True
+
+
+def _atexit():
+    if _show_run_warning and not "pytest" in sys.modules:
```
auscompgeek: lol. perhaps it'd be a better idea to move the pytest check down to the atexit register call?

wpilib/wpilib/_impl/main.py
```diff
@@ -11,6 +12,31 @@
...
+        print("")
+        print('    if __name__ == "__main__":')
+        print("        wpilib.run(MyRobot)")
+        print()
+
+
+def _excepthook(*args):
+    global _show_run_warning
+    _show_run_warning = False
+    _orig_excepthook(*args)
```
auscompgeek: maybe restore the original excepthook here too?

auscompgeek (commented):
how does this interact with someone playing around with wpilib in a repl?

wpilib/wpilib/_impl/main.py
virtuald: Eh.

virtuald:
One question is how poorly this would interact with something doing inspection of wpilib for autocomplete et al.

wpilib/wpilib/_impl/main.py
auscompgeek: Wouldn't this have a not-good impact on the time to run the tests though? Like, we have almost 900 tests, and wpilib gets reloaded for every one of those.

Issue #438, opened 2018-01-27
I2C OSError: Success


codetheweb:
I've been trying to get communication working over I2C between our roboRIO and an Arduino, and I've run into some issues.  It appears the Arduino (sometimes?) successfully receives data, but regardless of whether or not the data goes through an error is thrown on the roboRIO side.  

Here's the code:
```python
self.arduinoC = wpilib.I2C(wpilib.I2C.Port.kOnboard, 4)
self.arduinoC.transaction(b'sbt255', 1)
```
(docs for I2C are [here](http://robotpy.readthedocs.io/projects/wpilib/en/stable/wpilib/I2C.html?highlight=i2c#wpilib.i2c.I2C.transaction))

And here's the result of local tests:
```python
self = <hal_impl.i2c_helpers.I2CSimBase object at 0x10dc90d68>, port = 0, deviceAddress = 4, dataToSend = <hal.functions.c_ubyte_Array_6 object at 0x10dc6f488>, sendSize = 6
dataReceived = <hal.functions.c_ubyte_Array_1 object at 0x10dc6fbf8>, receiveSize = 1

    def transactionI2C(self, port, deviceAddress, dataToSend, sendSize, dataReceived, receiveSize):
        '''
                To give data back use ``data_received``::
    
                    data_received[:] = [1,2,3...]
    
                :returns: number of bytes returned
            '''
>       raise NotImplementedError
E       NotImplementedError
```

I'm guessing this is the result of 2018 changes to WPILib, but I'm not familiar enough with it to confirm.

virtuald:
What error shows up on the RoboRIO side? It'll be a different error I imagine. I don't know if we actually updated the I2C stuff for 2018, so that might be the issue. Seeing the error would be useful however.

When using I2C in simulation/testing, you have to implement the device side of the simulation (which is why you're getting a NotImplementedError). We don't really have a document yet describing how you would go about this... but I could put some documentation together tonight.

codetheweb:
Ok, I didn't realize the error was a result of it being in a simulation.

Unfortunately I won't have access to the robot until Monday, so I'll update you then.

virtuald:
I added a bunch of documentation, and the error messages will be more helpful in future releases. Hope that helps!

virtuald:
You can find it at http://robotpy.readthedocs.io/projects/wpilib/en/latest/hal.html#i2c-spi-simulation-helpers

codetheweb:
[...]

PR #580, opened 2019-02-01
hal-roborio: Fix getting errno

auscompgeek:
Fixes: #438


virtuald:
You said you tried this on a roborio? Looks good to me.

PR #583, opened 2019-02-01
Fix use of CANSparkMax in tests

auscompgeek:
We really need a way to let other packages hook into this...


virtuald:
I'm not a fan of this approach, I'd rather it be in the CAN dictionary. Maybe we change it to be 'spark-N'?

auscompgeek:
By that logic, the PDP and PCMs should also go in the CAN dictionary.

auscompgeek:
Closing in favour of robotpy/robotpy-rev#16, per above.

PR #588, opened 2019-02-13
Allow sendable chooser to respond to local NT changes

virtuald:
- Adds python-specific API to SendableBuilder
- Fixes https://github.com/robotpy/pyfrc/issues/155
- Fixes https://github.com/robotpy/robotpy-wpilib-utilities/issues/129


auscompgeek:
I think we need to be _really_ careful about exposing this sort of thing. I think the ramifications of setting this on a property with both a getter and a setter will be bad.

virtuald:
Noted, but this seems like the least-worst way to fix this.

PR #592, opened 2019-02-21
update max number of encoders available in hal-sim

jcaselman:
Limiting the number of encoders to the number of labeled DIO ports on the RIO divided by 2 causes failure when tests run using hal-sim

I believe the max number of encoders is number of DIO divided by 2, so this commit allows up to 13 encoders before an error is raised.


virtuald:
Thanks for catching this... though that has to be a *lot* of wiring for more than 5 encoders! 😲 

At some point it was 5, but as @auscompgeek points out the correct number is now 8. Please update your PR and we can merge this.

jcaselman:
Thanks for the quick response and feedback.  See latest commit.

PR #595, opened 2019-02-27
When simulation error occurs, raise a special exception to indicate it

virtuald:


hal-sim/hal_impl/fndef.py
```diff
@@ -206,20 +212,27 @@ def gen_func(f, name, restype, params, out, _thunk):
...
-            %s
-            return return_value
+            try:
+                %s
+                return_value = %s(%s)
+                %s
+            except NotImplementedError:
+                raise
+            except Exception as e:
+                raise SimulationError("unexpected exception calling '%s'") from e
```
auscompgeek: Do you think we should also pass through any HALErrors we get?

virtuald: I suppose, though they aren't used in simulation anywhere other than here.

PR #598, opened 2019-03-01
Make determining robot mode more efficient

virtuald:
We experience a lot of lock contention/overhead when calling these functions, so it makes sense to combine them. Upstream request at https://github.com/wpilibsuite/allwpilib/issues/1620

These are important, but pretty sure I got it right.


virtuald:
It's difficult to say accurately, but it's at least 4*function_call+1*ctypes_function_call+mutex_acquire+mutex_release.

PR #607, opened 2019-07-05
Add CPython 3.8 beta to CI matrix

auscompgeek:


auscompgeek:
Poked the pcpp author about this failure: ned14/pcpp#31

Issue #606, opened 2019-07-01
Drop Python 3.5 support?


auscompgeek:
Python 3.6 has variable annotations, which would be super-helpful for type hinting data classes.

PyPy3 has 3.6 support now (with a beta label - I hear coverage.py is currently a bit broken), so not too much of a concern there in case anyone (we?) wanted to go down that road.

In terms of distro support:
- Debian 9 (Stretch) is 3.5, but 10 (Buster) will be released in the next week with 3.7.
- The only supported Ubuntu release still on 3.5 is 16.04 (Xenial).
- RHEL7 has Python 3.6, of all distros...
- All supported Fedora releases are on 3.7.
- openSUSE Leap 42.3 has 3.4, which we don't support anyway. Leap 15.0-15.1 is 3.6.

virtuald:
I'm supportive of this.

auscompgeek:
I guess someone should make the PRs to remove 3.5 from CI. I'd like to see the magicbot examples updated to use the variable annotation syntax.

M-Shadow:
Think that should be about it. Any other repos that need updating to CI/updating in general based on this?

virtuald:
pyfrc, pynetworktables, navx, rev... we have a lot of repos. 

FWIW, robotpy-build uses f-strings, so that's definitely going to be 3.6 only.

auscompgeek:
We should also make sure the setup.py files list the 3.8 Trove classifier and not the 3.5 one as well.

virtuald:
Done.

PR #613, opened 2019-10-18
Change minimum required version to Python 3.6

M-Shadow:
Continuation of #612
Based on #606

### DIFF
--- a/wpilib/setup.py
+++ b/wpilib/setup.py
@@ -5,12 +5,6 @@ from os.path import dirname, exists, join
 import sys, subprocess
 from setuptools import find_packages, setup
 
-if sys.version_info.major < 3 or (
-    sys.version_info.major == 3 and sys.version_info.minor < 6
-):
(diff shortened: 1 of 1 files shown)


auscompgeek:
Might be worth taking the time to clean up this condition too whilst we're here.

M-Shadow:
Pretty sure we can just use the `python_requires=">=3.6"` setup() arg and just remove the conditional.

auscompgeek:
Yeah, I can't say I was ever really sure why we didn't just use `python_requires` here. We expect teams to install the pyfrc package on their computers after all, instead of installing wpilib directly.

On the other hand it does make some amount of sense for pyfrc to catch super-old versions of pip I guess.

virtuald:
I believe when I first created these things, I didn't realize that python_requires existed or (more likely) it didn't exist at the time.

auscompgeek:
python_requires is pretty old now; iirc it definitely existed at the time.

PR #629, opened 2020-02-06
Make Color and Color8Bit read-only and hashable

auscompgeek:


virtuald:
It would be nice if the builtin colors were readonly and the normal colors were read-write, but it does seem dangerous to allow users to modify the builtins.

auscompgeek:
I think I would prefer Color to be hashable, as it would be useful when using REV's ColorMatch.

PR #636, opened 2020-02-08
Simplify geometry objects

auscompgeek:
These are value types which are never passed around through shared_ptr.

Also reduce the redundancy of the signature for Rotation2d.fromDegrees.

### DIFF
--- a/gen/geometry/Pose2d.yml
+++ b/gen/geometry/Pose2d.yml
@@ -7,6 +7,7 @@ functions:
     ignore: true
 classes:
   Pose2d:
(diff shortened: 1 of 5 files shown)


gen/geometry/Pose2d.yml
```diff
@@ -7,7 +7,6 @@ functions:
     ignore: true
 classes:
   Pose2d:
-    shared_ptr: true
```
virtuald: shared_ptr defaults to true, so you should probably set it to false instead of deleting it...

PR #642, opened 2020-02-09
Ping pyfrc on successful publish

TheTripleV:
Relates to robotpy/pyfrc#165

Sends pyfrc a webhook to update it's wpilib dependency to the latest version.


TheTripleV:
This action needs a repo access token in `secrets.REPO_ACCESS_TOKEN`

TheTripleV:
This now sends pyfrc its version number. Note that the code for this only works if `startsWith(github.event.ref, 'refs/tags')`. Otherwise, we would have to clone the repo.

PR #643, opened 2020-02-10
Fix feedforward classes to be unitless

auscompgeek:


virtuald:
I'm not sure this is a good idea, because the units interact with volts (kv, ka) internally, unlike the controller classes.

auscompgeek:
Sure, but what units do you use to measure the distance a flywheel spins?

The Java API makes these unitless because the units do indeed cancel out.

virtuald:
I would be fine with you adding a scalar version that has the original name. I don't really want to remove them at this time, because we've already broken users enough times this season.

wpilib/controller/__init__.py
```diff
@@ -3,22 +3,22 @@
...
     ProfiledPIDController,
     RamseteController,
+    SimpleMotorFeedforward,
     SimpleMotorFeedforwardFeet,
     SimpleMotorFeedforwardMeters,
 )
 
 __all__ = [
     "ArmFeedforward",
-    "ElevatorFeedforwardFeet",
```
virtuald: Should still be in all.

auscompgeek: Do we really care about documenting these? I don't think we want to actively encourage users to use them if we're going to remove them next season.

PR #657, opened 2020-02-16
Hide Filesystem.h, it's useless from Python

auscompgeek:


virtuald:
Seems harmless?

auscompgeek:
I mean I'm sure someone will get confused when they see this type: https://robotpy.readthedocs.io/projects/wpilib/en/latest/wpilib/functions.html#wpilib.getDeployDirectory

PR #665, opened 2020-02-19
Make Translation2d more Pythonic

auscompgeek:


virtuald:
I'm not so sure about making it like a tuple. What about the other geometry classes then?

PR #668, opened 2020-02-24
Add feet to Pose2d/Translation2d

virtuald:
@auscompgeek let's fight about this. I'll go first.

... you're a poopy head!

### DIFF
--- a/gen/geometry/Transform2d.yml
+++ b/gen/geometry/Transform2d.yml
@@ -6,6 +6,8 @@ extra_includes:
 classes:
   Transform2d:
     shared_ptr: false
+    force_type_casters:
+    - units::unit_t
     methods:
       Transform2d:
         overloads:
@@ -17,3 +19,13 @@ classes:
       operator*:
       operator==:
       operator!=:
+
+inline_code: |
+  cls_Transform2d
+    .def_static("fromFeet", [](units::foot_t x, units::foot_t y, units::radian_t angle){
+      return Transform2d(Translation2d(x, y), Rotation2d(angle));
+    }, py::arg("x"), py::arg("y"), py::arg("angle"))
+    .def(py::init([](units::meter_t x, units::meter_t y, units::radian_t angle) {
+      return Transform2d(Translation2d(x, y), Rotation2d(angle));
+    }), py::arg("x"), py::arg("y"), py::arg("angle"))
+  ;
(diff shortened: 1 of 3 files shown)


gen/geometry/Translation2d.yml
```diff
@@ -36,8 +36,23 @@ classes:
...
+      return Translation2d(x, y);
+    })
     .def_property_readonly("x", &Translation2d::X)
     .def_property_readonly("y", &Translation2d::Y)
+    .def_property_readonly("x_feet", [](Translation2d * self) -> units::foot_t {
+      return self->X();
+    })
+    .def_property_readonly("y_feet", [](Translation2d * self) -> units::foot_t {
+      return self->Y();
+    })
```
auscompgeek: These seem odd. Surely they should be getter methods instead?

gen/geometry/Transform2d.yml
```diff
@@ -17,3 +19,13 @@ classes:
...
+
+inline_code: |
+  cls_Transform2d
+    .def_static("fromFeet", [](units::foot_t x, units::foot_t y, units::radian_t angle){
+      return Transform2d(Translation2d(x, y), Rotation2d(angle));
+    }, py::arg("x"), py::arg("y"), py::arg("angle"))
+    .def(py::init([](units::meter_t x, units::meter_t y, units::radian_t angle) {
+      return Transform2d(Translation2d(x, y), Rotation2d(angle));
+    }), py::arg("x"), py::arg("y"), py::arg("angle"))
+  ;
```
auscompgeek: I'm confused. How often would someone actually want to construct a Transform2d themselves?

gen/geometry/Translation2d.yml
virtuald: They're in the same vein as x, so provided for consistency.

gen/geometry/Transform2d.yml
virtuald: I'm using them in pyfrc physics updates.

virtuald:
I don't really like having separate conversion functions, they're a lot more verbose and they require more overhead because you're dealing with python code doing the conversion/validation (whereas the C++ conversions are basically free).

gen/geometry/Translation2d.yml
auscompgeek: Ehh. These convert though, the others don't. I really only added those because the method names are rather silly and I didn't want to break anyone who was already using them.

PR #673, opened 2020-03-01
Fix geometry/kinematics reprs

auscompgeek:
[From the Python documentation for `repr()`](https://docs.python.org/3/library/functions.html#repr):

> For many types, this function makes an attempt to return a string that would yield an object with the same value when passed to `eval()`, otherwise the representation is a string enclosed in angle brackets [...]

This reverts the geometry `__repr__` and fixes the kinematics `__repr__` implementations to adhere to the expected behaviour of `repr()`.

### DIFF
--- a/gen/kinematics/ChassisSpeeds.yml
+++ b/gen/kinematics/ChassisSpeeds.yml
@@ -48,8 +48,8 @@ inline_code: |
       }
     )
     .def("__repr__", [](const ChassisSpeeds &cs) -> std::string {
-      return "ChassisSpeeds(vx=" + std::to_string(cs.vx()) + "m/s, "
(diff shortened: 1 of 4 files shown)


virtuald:
I would rather have the units present, to remind users that these objects have units associated with them and that they're not just raw values.

auscompgeek:
I would much rather be able to recreate these by just copy-pasting. You don't see the units when you instantiate these anyway.

The Java toString implementations don't include units either.

virtuald:
How often are you recreating these?

auscompgeek:
look I just know I'm going to screw up the maths when converting between our robot field coordinate system and the simulation coordinate system :P

PR #687, opened 2021-02-15
Fix empty sendable chooser

virtuald:
Demonstration of prior error:

```
    def test_sendable_chooser():
        chooser = wpilib.SendableChooser()
>       assert chooser.getSelected() is None
E       TypeError: Unable to convert function return value to a Python type! The signature was
E               (self: wpilib._wpilib.SendableChooser) -> object
```

### DIFF
--- a/gen/SendableChooser.yml
+++ b/gen/SendableChooser.yml
@@ -15,8 +15,12 @@ classes:
       GetSelected:
         # weirdness because return type
         cpp_code: |
-          [](frc::SendableChooser<T> * __that) {
-            return __that->GetSelected();
+          [](frc::SendableChooser<T> * __that) -> py::object {
+            auto v = __that->GetSelected();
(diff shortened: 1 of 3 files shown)


gen/SendableChooser.yml
```diff
@@ -15,8 +15,12 @@ classes:
       GetSelected:
         # weirdness because return type
         cpp_code: |
-          [](frc::SendableChooser<T> * __that) {
-            return __that->GetSelected();
+          [](frc::SendableChooser<T> * __that) -> py::object {
+            auto v = __that->GetSelected();
+            if (!v) {
```
auscompgeek, line 20: What if the selection is False or 0?

virtuald: The operator bool is for checking if the handle is valid or not.

PR #690, opened 2021-02-17
Suppress CTRL-C segfaults on Linux

### LINKED ISSUE #689: CTRL-C in gui on Linux segfaults
I've been playing with this but I think I give up. If I CTRL-C the gui, it doesn't do anything until I close the window, at which point this shows up:

```
^CTraceback (most recent call last):
  File "/home/virtuald/src/frc/examples/commands-v2/ramsete/./robot.py", line 65, in <module>
    wpilib.run(MyRobot)
  File "/home/virtuald/src/frc/robotpy-wpilib/wpilib/_impl/main.py", line 193, in run
    retval = options.cmdobj.run(options, robot_class, **kwargs)
  File "/home/virtuald/src/frc/pyfrc/pyfrc/mains/cli_sim.py", line 80, in run
    return robot_class.main(robot_class)
  File "/home/virtuald/src/frc/robotpy-wpilib/wpilib/_impl/start.py", line 37, in run
    hal.runMain()
KeyboardInterrupt
terminate called without an active exception
Fatal Python error: Aborted

Thread 0x00007fb8ad1ce740 (most recent call first):
<no Python frame>
Aborted (core dumped)
```

I had thought maybe the right way to deal with it is add the following to robotpy-hal:

```
  // add a simulation periodic handler that checks to see if python needs
  // to handle any signals. Only done in simulation mode.
  HALSIM_RegisterSimPeriodicBeforeCallback(
      [](void *) {
        py::gil_scoped_acquire gil;
        if (PyErr_CheckSignals()) {
          throw py::error_already_set();
        }
      },
      nullptr);
```

It seems that PyErr_CheckSignals never returned non-zero. I played with a variety of things, but nothing really stuck.

I did add a signal handler right before `hal.runMain()` is called, and that prevents the segfault from occurring, but the application still doesn't end. Interestingly enough, the signal handler doesn't get called until after the window is closed. 

```
def _on_ctrl_c(*ignored):
    hal.exitMain()

signal.signal(signal.SIGINT, _on_ctrl_c)
```

I guess it does solve the segfault, but it feels like a bad idea. Not sure what happens on Windows/OSX.

virtuald:
Fixes #689

@TheTripleV can you try this on Windows? Curious what this does there if you CTRL-C when the gui is running.

### DIFF
--- a/wpilib/_impl/start.py
+++ b/wpilib/_impl/start.py
@@ -1,6 +1,7 @@
 import hal
 import wpilib
 import logging
+import signal
 import threading
 
 
@@ -22,6 +23,9 @@ class RobotStarter:
         self.robot = None
         self.suppressExitWarning = False
 
+    def _on_ctrl_c(self, *ignored):
+        hal.exitMain()
+
     def run(self, robot_cls: wpilib.RobotBase) -> bool:
         if hal.hasMain():
             rval = [False]
@@ -31,6 +35,8 @@ class RobotStarter:
                     rval[0] = self.start(robot_cls)
                 finally:
                     hal.exitMain()
+
+            signal.signal(signal.SIGINT, self._on_ctrl_c)
 
             th = threading.Thread(target=_start, name="RobotThread", daemon=True)
             th.start()


wpilib/_impl/start.py
```diff
@@ -32,6 +36,8 @@ def _start():
                 finally:
                     hal.exitMain()
 
+            signal.signal(signal.SIGINT, self._on_ctrl_c)
```
auscompgeek, line 39: Wouldn't it be better to wrap the runMain in a try/except?

TheTripleV:
Ctrl-C does nothing, running the sim both with and without the gui.

wpilib/_impl/start.py
virtuald: I don't see why it matters in this case.

virtuald:
OBE, fixed in halsim-gui instead.

PR #699, opened 2022-01-26
Fix DifferentialDrive docs examples

etiennebeaulac:
The current `DifferentialDrive` docs examples do not work since `DifferentialDrive` is located in the `wpilib.drive` module.

I also added the two separate imports that are needed : the second is often forgotten by new Python students.

### DIFF
--- a/gen/drive/DifferentialDrive.yml
+++ b/gen/drive/DifferentialDrive.yml
@@ -41,6 +41,8 @@ classes:
       :class:`MotorControllerGroup` instances as follows.
 
       Four motor drivetrain::
+        import wpilib
+        import wpilib.drive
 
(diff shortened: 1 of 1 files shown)


virtuald:
Thanks!

gen/drive/DifferentialDrive.yml
```diff
@@ -41,6 +41,8 @@ classes:
       :class:`MotorControllerGroup` instances as follows.
 
       Four motor drivetrain::
+        import wpilib
+        import wpilib.drive
```
auscompgeek, line 46: FYI, blank lines are significant in reStructuredText. I've opened a PR: #700

PR #700, opened 2022-01-26
Fix DifferentialDrive code example rendering

auscompgeek:
Before:

image

### DIFF
--- a/gen/drive/DifferentialDrive.yml
+++ b/gen/drive/DifferentialDrive.yml
@@ -41,7 +41,7 @@ classes:
       :class:`MotorControllerGroup` instances as follows.
 
       Four motor drivetrain::
-        import wpilib
+
         import wpilib.drive
 
         class Robot(wpilib.TimedRobot):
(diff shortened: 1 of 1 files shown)


gen/drive/DifferentialDrive.yml
```diff
@@ -57,7 +57,7 @@ classes:
                 self.drive = wpilib.drive.DifferentialDrive(self.left, self.right)
 
       Six motor drivetrain::
-        import wpilib
+
```
virtuald, line 60: The example needs `import wpilib` to work?

auscompgeek: It's already imported on the line below.

virtuald: ... heh, since when does that work?

Well, put it back anyways. If you don't import wpilib separately, autocomplete breaks.

auscompgeek: > autocomplete breaks

It works fine with Jedi for me. What IDE/configuration are you using?

virtuald: vscode

virtuald: fine I don't care

PR #703, opened 2022-02-04
Prevent abort() on CTRL-C on sim on Linux

virtuald:
- This isn't quite right, something is handling the signal and the CTRL-C isn't getting processed somehow


virtuald:
Combined with https://github.com/robotpy/robotpy-halsim-gui/pull/10 we achieve expected behavior.

PR #707, opened 2022-03-04
Display deploy.json data at robot startup

virtuald:
Not quite sure if I like the naming or placement of these. Open for bikeshedding. Currently looks like this:

```
02:17:53:468 INFO    : deploy-info         : virtuald@hostname at 2022-03-04T02:04:09
02:17:53:468 INFO    : deploy-info         : - git info: fa1331a-dirty (branch=main)
```

### DIFF
--- a/wpilib/deployinfo.py
+++ b/wpilib/deployinfo.py
@@ -1,0 +1,33 @@
+from ._wpilib import RobotBase
+import json
+import typing
+
+
+def getDeployData() -> typing.Union[None, typing.Dict[str, str]]:
+    """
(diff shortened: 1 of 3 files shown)


wpilib/deployinfo.py
```diff
@@ -0,0 +1,33 @@
+from ._wpilib import RobotBase
+import json
+import typing
+
+
+def getDeployData() -> typing.Union[None, typing.Dict[str, str]]:
```
auscompgeek, line 6:
```suggestion
def getDeployData() -> typing.Optional[typing.Dict[str, str]]:
```

Issue #691, opened 2021-02-22
wpilib.shuffleboard not implemented


virtuald:
There's a virtual base class in there somewhere which made this tricky to do correctly, though I don't quite remember why that is.

virtuald:
I have something for this, but it needs testing.

auscompgeek:
We currently have a TODO comment:

https://github.com/robotpy/robotpy-wpilib/blob/69cf7d37c1bff7d278464a091ced1a6154ec0ba0/pyproject.toml#L282

Reopening to track this.

auscompgeek:
@virtuald do you recall how exactly it's broken? I presume it's not related to https://github.com/wpilibsuite/BetaTest/issues/80

virtuald:
I feel it was some NT4 thing, but I imagine that's been fixed by now. We should re-enable and add a simple test to make sure it works.

PR #715, opened 2022-11-12
2023 beta

virtuald:


gen/shuffleboard/ShuffleboardContainer.yml
```diff
@@ -6,66 +6,72 @@ extra_includes:
...
+          std::string_view, std::string_view:
+          std::string_view, char*:
+            ignore: true
+          std::string_view, std::span<const bool>:
+          std::string_view, std::span<const double>:
+          std::string_view, std::span<const float>:
+          std::string_view, std::span<const int64_t>:
+          std::string_view, std::span<const std::string>:
+      AddCamera:
+        ignore: true
```
auscompgeek, line 46: Is there a reason this is ignored?

virtuald: You can't have cscore loaded in the same process as everything else otherwise bad things happen

auscompgeek: This method shouldn't depend on cscore?

virtuald: Oh, you're right. Oops.

... all of shuffleboard is disabled at the moment since I didn't have time to re-enable it. It might work?

auscompgeek: Right, fair enough. That means #691 was prematurely closed, right?

PR #717, opened 2022-12-31
Fix shuffleboard package

virtuald:
- Lots of places where reference_internal is needed because shuffleboard uses references extensively
- Add somewhere to hold wpi::Sendables just like SmartDashboard. Can't reuse that because they're in different packages

Blocked on https://github.com/wpilibsuite/allwpilib/pull/4884


gen/shuffleboard/ShuffleboardContainer.yml
```diff
@@ -14,64 +17,148 @@ classes:
...
           std::string_view, std::span<const double>:
+            return_value_policy: reference_internal
           std::string_view, std::span<const float>:
+            return_value_policy: reference_internal
           std::string_view, std::span<const int64_t>:
+            return_value_policy: reference_internal
           std::string_view, std::span<const std::string>:
+            return_value_policy: reference_internal
       AddCamera:
         ignore: true
```
auscompgeek, line 108:
```suggestion
```

gen/shuffleboard/ShuffleboardContainer.yml
```diff
@@ -14,64 +17,147 @@ classes:
...
+            param_override:
+              sendable:
+                name: defaultValue
+            cpp_code: |
+              [](ShuffleboardContainer *self, py::str &key, std::shared_ptr<wpi::Sendable> sendable) -> frc::ComplexWidget& {
+                if (!sendable) {
+                  throw FRC_MakeError(err::NullParameter, "{}", "value");
+                }
+
+                // convert key to a raw string so that we can create a StringRef
```
auscompgeek, line 46: Doesn't pybind11 support string_view arguments?

virtuald: Yes, but I need to store a py::str in the shuffeboard data anyways, so this makes it so I don't have to recreate it.

gen/shuffleboard/ShuffleboardContainer.yml
```diff
@@ -14,64 +17,148 @@ classes:
...
           std::string_view, int:
+            return_value_policy: reference_internal
           std::string_view, std::string_view:
+            return_value_policy: reference_internal
           std::string_view, std::span<const bool>:
+            return_value_policy: reference_internal
           std::string_view, std::span<const double>:
+            return_value_policy: reference_internal
           std::string_view, std::span<const float>:
+            return_value_policy: reference_internal
```
auscompgeek, line 158: Will this overload be reachable? Maybe passing in a numpy float32 array could work (I haven't tried), but will it hit the `std::span<const double>` type caster before reaching this?

virtuald: Probably not. But really, will anyone care?

auscompgeek: It'll probably be confusing if we leave an unreachable thing documented.

virtuald: Well when someone tries to reach it they can file a bug, and then we know someone cares about it.

PR #719, opened 2023-01-05
Disarm GIL threadstate deletion if interpreter is exiting

virtuald:


virtuald:
I'm not really sure this is the right fix, but it'll do for now.

virtuald:
This did not actually solve the problem, something else is happening.

auscompgeek:
Maybe SafeThread is blowing up?

virtuald:
Yes, that was also an issue. I forgot to push a fix for that last night.

